PR #62, opened 2012-08-15
a new pull request

rgraneru:
removed that netbeans file. And then in my next commit i downgraded some debug logging to trace


jwestberg:
Looks good!

PR #73, opened 2012-09-14
Moved the stages created for a customer to the common repo for everyone to enjoy. Also extended the functionality of the RexExp stage

rgraneru:
The stages are:

FilesToDiskDumber: Dumps the contents of a binary file to the disk in a readable format. For debugging purposes.

RemoveFields: Remove specific fields from Hydra

PrioritizedCopyStage: A advanced copystage. Specify a number of fields and it will copy the contents of the first of those fields that exist into a new field. It also supports adding a condition for the copy to be done. This stage will not overwrite any existings fields, except if you use the overWrite variable.

PopulateQueryCompletionFields: Another copy-stage that will check in a specific field and create new fields based on what it finds. For example: If it finds foo and bar in the collection fields, it would create foo_qc and bar_qc with the contents of the copyFields. It also supports renaming the fieldnames found in the collection fields so that when foo is found it creates superfoo_qc

HtmlUnescaperStage: A stage that will unescape html characters.


laserval:
Personally, I'd prefer a more generic name for PopulateQueryCompletionFields, since QC has nothing to do with it's function. From what I understand, it is a mapping stage that maps the content of a number of fields (defined in a separate field) to one or more fields that have a suffix appended to their field names. So basically it is a CopyStage where the document can control what fields should be copied.
Perhaps it can be made more general to reflect that?
If I misunderstood it, please correct.

EDIT: Ok, I realise it's the other way around - the document controls which fields to copy to, not from. And that's also what you said in the pull request, bleh :P

rgraneru:
You understand correctly and I agree with you. But now we end up with
so many copy-stages so naming them is a bit difficult.  And merging
them is a bit difficult too. But I'll think about it and think about
it.

Thanks for the feedback

laserval:
After some discussions with @jwestberg , I think it's best to close this pull request so that you can submit separate ones for some of the stages. It is pretty big and many of the changes are unrelated.

Also mostly because of this:
- `HtmlUnescaperStage` depends on springframework, which will make the basic-stages jar balloon in size. It might also fit better in web-stages, but that is debatable :)
- `PopulateQueryCompletionFields` should be renamed/refactored

If you can submit the Regex changes, RemoveFields and the `PrioritizedCopyStage` in a pull request those could be merged I think.

rgraneru:
No problem. I'll do that

Roar

PR #82, opened 2012-09-20
Improved StageRunner, fixes #81

### LINKED ISSUE #81: Windows: StageRunner restarts stage that was killed
It looks like the exitValue from a killed process in Windows (tested Windows 7, Windows 2008 Server) is not 143, as is expected. 

Exit code appears to be 1 instead. When stage configuration is modified, a Hydra core running on Windows will start up a stage in the new configuration while simultaneously restarting the old one.

jwestberg:
Simply added a flag which lets the restarting logic know if the stage was killed on purpose or not.


laserval:
I'm not sure I understand how this works. Where is the flag set to anything other than false?

jwestberg:
Oops. The line with the mysterious missing assignment is now added.

PR #90, opened 2012-09-28
Fixed StageRunner for Windows

jwestberg:
Stages did not properly restart, seemingly at random in Windows. 

Long running tests using this have not revealed any issues with crashed stages not restarting on windows.


jwestberg:
Tested on ~10000 documents on Windows prior to and after this fix. Before fix, the pipeline would get stuck on non-restarting stages 3-4 times. After the fix, I've had no problems.

PR #92, opened 2012-10-08
Groovy runner stage

antonha:
Figured that stages could be more easily managed if all the runtime of a stage could be passed along as arguments representable as Strings.

Thus, I implemented a Stage which takes a Groovy string, tries to instantiate a class from that file (which extends a certain interface) and then tries to run that script on the LocalDocument passed in process().

Still in very-very-much beta, this can hopefully make it easier to add new stages to an existing pipeline. Together with an admin-interface, this can get interesting!


laserval:
Cool stuff!

Wondering about where in the directory structure this should be though. Perhaps we need a /stages/runners directory. Could you move this into a fitting place in the stages dir?

antonha:
Moved it to stages/processing, since the stage actually extends the AbstractProcessingStage.

laserval:
Well, AbstractProcessStage is the base for all stages except InputStages so it's not a basis for placing it in processing. But the directory structure might need some consideration anyway.

laserval:
Missed that there are a lack of annotations. If you can submit a fix with the proper annotations on the GroovyRunnerStage that'd be swell.

PR #99, opened 2012-10-12
Adding numeration of target fields when url field is multi-valued

jwestberg:
Also fixed a broken test in SimpleFetchingTikaStage which rather confusingly killed any subsequent test due to partial mocks in Mockito.


jwestberg:
Added proper URI-encoding in TikaUtils as well.

jwestberg:
Also adding language detection, while I'm at it :)

rgraneru:
I've started testing the new code and will merge it when I'm done.

PR #100, opened 2012-10-15
Added support for fetching basic document information for the admin service

simonstenstrom:
- Added two endpoints /documents/count and /documents/list 
- The first returns the number of documents matching a query (by adding for example ?q={fetched:true})
- The second returns the documents matching a query (by adding for example ?q={fetched:true}). You can also specify the number of documents to return (by adding for example &limit=10) and skip a number of documents for pagination (by adding for example &skip=10)
- I no query is specified, all documents are considered matches.

### DIFF
--- a/api/src/main/java/com/findwise/hydra/local/LocalQuery.java
+++ b/api/src/main/java/com/findwise/hydra/local/LocalQuery.java
@@ -12,12 +12,14 @@ import com.google.gson.JsonParseException;
 
 public class LocalQuery implements Query, JsonDeserializer {
 	private Map<String, Object> equals;
+	private Map<String, Object> notEquals;
 	private Map<String, Boolean> exists;
 	private Map<String, Boolean> touched;
 	private Action action = null;
 	
 	public LocalQuery() {
 		equals = new HashMap<String, Object>();
+		notEquals = new HashMap<String, Object>();
 		exists = new HashMap<String, Boolean>();
 		touched = new HashMap<String, Boolean>();
 	}
@@ -77,6 +79,7 @@ public class LocalQuery implements Query, JsonDeserializer {
 	public String toJson() {
 		Map<String, Object> x = new HashMap<String, Object>();
 		x.put("equals", equals);
+		x.put("notEquals", notEquals);
 		x.put("exists", exists);
 		x.put("touched", touched);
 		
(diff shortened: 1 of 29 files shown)


api/src/main/java/com/findwise/hydra/local/LocalQuery.java
```diff
@@ -115,4 +121,41 @@ public void fromJson(String json) throws JsonException{
...
+
+	public void setEquals(Map<String, Object> equals) {
+		this.equals = equals;
+	}
+
+	public Map<String, Object> getNotEquals() {
+		return notEquals;
+	}
+
+	public void setNotEquals(Map<String, Object> notEquals) {
```
jwestberg: Shouldn't this be modified by reference instead? Why would you ever need to setNotEquals()? getNotEquals should never return null, so it should really just be modified (changed by reference).

database-impl/mongodb/src/main/java/com/findwise/hydra/mongodb/MongoConnector.java
```diff
@@ -222,4 +224,23 @@ public MongoStatusIO getStatusWriter() {
 	public MongoStatusIO getStatusReader() {
 		return statusIO;
 	}
+
+	@Override
+	public DatabaseDocument<MongoType> convert(DatabaseDocument<?> document) {
```
jwestberg: We could probably merge this convert(DatabaseDocument<?> and the convert(LocalDocument) to a simple convert(Document), no? Since Document extends the JsonDeserializer interface, this shouldn't be a big problem.

Seems easier to do that than to commit a bunch of Database-API changes and implement those all over :)

database/src/main/java/com/findwise/hydra/DatabaseConnector.java
```diff
@@ -36,9 +36,12 @@
 	DocumentWriter<T> getDocumentWriter();
 	
 	DatabaseQuery<T> convert(LocalQuery query);
+	DatabaseQuery<T> convert(DatabaseQuery<?> query);
```
jwestberg: -38
-39
- DatabaseQuery<T> convert(Query query); ? :)

jwestberg: -38
-39
- DatabaseQuery<T> convert(Query query); ? :)

api/src/main/java/com/findwise/hydra/local/LocalQuery.java
simonstenstrom: Agreed. Changes pushed

simonstenstrom:
Fixes committed

simonstenstrom:
Rebased done

simonstenstrom:
Added support for monitoring stages and adding/updating stage configuration. Also updated readme file with documentation on how the admin service should be used for the new features

simonstenstrom:
Any progress with this merge?

rgraneru:
Sorry about the delay. I plan to test it tomorrow.

PR #101, opened 2012-10-19
Adding support for fields containing lists

jwestberg:
This obsoletes the MergeListsStage as this more generically solves the same basic problem. I think.


rgraneru:
How is this going? Is the pull request amended yet?

jwestberg:
Done now.

PR #104, opened 2012-10-25
JVM selection

jwestberg:
Adding support for specifying a JVM installation for a stage by specifying java_location property.

Also, made RESTServer a Daemon to ensure Hydra process exits properly if NodeMaster fails to start (usually due to Mongo not being available)

### DIFF
--- a/core/src/main/java/com/findwise/hydra/StageRunner.java
+++ b/core/src/main/java/com/findwise/hydra/StageRunner.java
@@ -28,6 +28,7 @@ public class StageRunner extends Thread {
     private int timesStarted;
     private boolean loggingEnabled;
     private String jvmParameters;
+    private String java;
     private String startupArgsString;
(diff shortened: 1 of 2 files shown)


rgraneru:
Do you know how this works in windows? As a Daemon I mean.

jwestberg:
Apples and oranges Roar... http://docs.oracle.com/javase/6/docs/api/java/lang/Thread.html#setDaemon(boolean)

rgraneru:
You meant that the os and the JVM are like apples and oranges, right?

Or windows and *nix

jwestberg:
Yeah, this has nothing to do with the status of the Hydra process as a daemon/service or not. As it says in the API docs I linked, this is a flag on the Thread instance to indicate it is not a user thread (e.g. blocks termination of the jvm)...

PR #106, opened 2012-10-29
Implementation of Stage Groups

jwestberg:
With the merge of this pull request, it becomes possible to configure hydra stages in groups. All stages in a StageGroup will be run in the same JVM, allowing much greater control of the processes in play in your Pipeline.

A `Pipeline` now contains only `StageGroups` which in turn contain the actual `Stage` objects. As a result, configuring via the Database packages has changed, but old configurations in MongoDB will still function (upon reading a stage that does not belong to a group, a new group matching the name of the stage will be added by the MongoPipelineReader).

StageGroups can contain configurations on the JVM level - such as the jvm installation directory, classpath information etc - while all other configuration stays in the configuration of the specific stage. 

Tested to be working within the same jar file as well as with multiple jars :)


rgraneru:
Man, these pull requests are big. I'm not even sure where to start checking that this doesn't break anything important.

jwestberg:
Well, it's a major feature addition in the central part of a large and complex piece of software. I assure you, it works though, and the code is actually much better than it used to be as well for a lot of classes in the hydra-core package.

There's a lot of unnecessary commits referenced though, just look at the files changed instead. And test it.

rgraneru:
I tried to merge this with https://github.com/Findwise/Hydra/pull/100 (simons Added support for fetching basic document information for the admin service)

I got a pretty big merge conflict with database/src/main/java/com/findwise/hydra/Pipeline.java

And since Simons pull request is the oldest I make this your "problem" :)

Try to pull 
https://github.com/simonstenstrom/Hydra DocOverview
into this branch and see if the merge conflict is as serious as it looks to me.

jwestberg:
Please merge #100 if that looks good as it is? 

I won't manually merge with another branch which is still open for modification, and thus have another conflict to handle as soon as that branch is patched and merged with master later on.

rgraneru:
will do!

rgraneru:
done now

jwestberg:
I gave up and did a hard merge rather than rebase. Tested and working properly.

PR #107, opened 2012-10-30
postfix/prefix conditions added to copyStage and prioritizedCopyStage

rgraneru:
It is not possible to add wildcard postfix and prefix arguments to copyField. That is you can tell it to copy ANYTHING_content to body, for example.  Or phone_ANYTHING to phone_number. 

Also removed some unused imports from basic stages.

### DIFF
--- a/stages/processing/basic/src/main/java/com/findwise/hydra/stage/CopyStage.java
+++ b/stages/processing/basic/src/main/java/com/findwise/hydra/stage/CopyStage.java
@@ -8,13 +8,15 @@ import com.findwise.hydra.local.LocalDocument;
  */
 @Stage(description = "Copies values from one field to another.")
 public class CopyStage extends AbstractMappingProcessStage {
+
     @Parameter(name = "prefix", description = "Input field prefix")
     private String prefix = "";
     @Parameter(name = "postfix", description = "Input field postfix")
     private String postfix = "";
-    
+
     @Override
-    public void stageInit() throws RequiredArgumentMissingException { }
+    public void stageInit() throws RequiredArgumentMissingException {
+    }
 
     @Override
     public void processField(LocalDocument doc, String fromField, String toField)
@@ -24,9 +26,9 @@ public class CopyStage extends AbstractMappingProcessStage {
         doc.putContentField(toField, val);
         String valString = (val == null) ? null : val.toString();
         Logger.debug("Copying field " + fromFieldStr + " to field " + toField
-            + " value: " + valString);
+                + " value: " + valString);
     }
-    
+
     public void setPostfix(String postfix) {
         this.postfix = postfix;
     }
--- a/stages/processing/basic/src/main/java/com/findwise/hydra/stage/CopyStage.java
+++ b/stages/processing/basic/src/main/java/com/findwise/hydra/stage/CopyStage.java
@@ -4,13 +4,13 @@ import com.findwise.hydra.common.Logger;
 import com.findwise.hydra.local.LocalDocument;
 
 /**
- * A basic Copy-stage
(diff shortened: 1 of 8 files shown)


stages/processing/basic/src/main/java/com/findwise/hydra/stage/CopyStage.java
```diff
@@ -4,20 +4,36 @@
...
+    private String postfix = "";
+
+    @Override
+    public void stageInit() throws RequiredArgumentMissingException {
+    }
+
+    @Override
+    public void processField(LocalDocument doc, String fromField, String toField)
+            throws ProcessException {
+        String fromFieldStr = prefix + fromField + postfix;
```
jwestberg, line 24: Why would you ever need this? Why not just configure the input field to be whatever the actual field name is?

rgraneru, line 24: Because you won't always know what these fields are. For example when you
use the Tika extraction stage to get the contents of a file, the default
name of the field containing the actual content will be
name_of_file_content, and the title  metadata field would be
name_of_file_title.

We could just make the tika extractor call the field the same thing for
every document ofc, but this is something that is good to have anyway. At
gard they kept changing the name of their binaryfield, so in the end we had
5 different fields we tried to copy to body. But they all ended in _content
so with this feature it would have been smoother.

Roar

jwestberg, line 24: I still don't see the difference. Suppose fromField is the sequence of `["x", "y", "z"]` with prefix `"pre_"` and postfix `"_post"`. You will invariably generate the fromFieldStr sequence of `["pre_x_post", "pre_y_post", "pre_z_post"]` right? 

So why not just use that as the fromField if you always need it to be that way?

rgraneru, line 24: Yes. I agree. Olof just spent 15 minutes explaining it to me.

Please don't accept this pull-request. It is just stupid. :)

rgraneru, line 24: Do you have any idea how I could implement the features I want.

I could like to simulate:

cp *_content body

and

cp document_\* body

The problem is if you got 2 or more matches.

laserval:
Well, LocalDocument doesn't have wildcard field gets. You'd have to get all field names and loop over them to find the ones that match your wildcard, or something like that.

Since this pull request doesn't do what you intended, I'll close it. Submit a new one when you have the full functionality implemented.

rgraneru:
Yes, do that. Wonderful.

And yes, I have to go through every one of them and check what matches
using regexp.

But should I create a new stage for this or expand copyField to allow you
to type for example:

fromfield: *_content
tofield: body

This will then ruin it if a field is called ***_SUPERFIELD**_\* and you
would like to copy it, but who would do such a thing? :)

jwestberg:
Are you sure you want this? I'm not at all certain what `*_content -> content` would mean if several fields are matched. Perhaps [MergeFieldsStage](https://github.com/Findwise/Hydra/blob/master/stages/processing/basic/src/main/java/com/findwise/hydra/stage/MergeFieldsStage.java) already does what you want? It allows merging of fields in a many-to-one manner with regex to describe the many. If it only matches exactly one field, it would more or less be a copy (for Strings, Lists, and Numbers anyway)

rgraneru:
That might be. I'll check it out

rgraneru:
I did check it out. And as always, you're right. It did what I wanted.
Thanks

PR #113, opened 2012-10-31
Annotations for the Groovy Runner Stage.

antonha:
In https://github.com/Findwise/Hydra/pull/92 there was a lack off annotations, this commit fixes that.


rgraneru:
When I pulled this and tried to build it, it failed because it tried to build as java 1.3. Adding a 1.6 force to the pom fixes the problem. But is this the correct solution to this problem?

```
     <plugin>
      <groupId>org.apache.maven.plugins</groupId>
      <artifactId>maven-compiler-plugin</artifactId>
      <version>2.3.2</version>
      <configuration>
       <source>1.6</source>
       <target>1.6</target>
      </configuration>
     </plugin>
```

antonha:
If the explicit setting of version is needed depends on your general maven setup. Probably a good idea to include it.

PR #115, opened 2012-11-02
Added a failsafe to solrOutput stage so that if you send in a document w...

rgraneru:
...here you haven't set action.ADD or action.DELETE, the stage will throw an exception instead of doing nothing

### DIFF
--- a/stages/out/solr-out/src/main/java/com/findwise/hydra/output/solr/SolrOutputStage.java
+++ b/stages/out/solr-out/src/main/java/com/findwise/hydra/output/solr/SolrOutputStage.java
@@ -46,6 +46,10 @@ public class SolrOutputStage extends AbstractOutputStage {
 			add(doc);
 		} else if (action == Action.DELETE) {
 			delete(doc);
+		} else if (action == Action.UPDATE) {
+			failDocument(doc, new RequiredArgumentMissingException("action UPDATE is not supported. Failing document"));
+		} else{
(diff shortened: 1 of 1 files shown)


stages/out/solr-out/src/main/java/com/findwise/hydra/output/solr/SolrOutputStage.java
```diff
@@ -46,6 +46,10 @@ public void output(LocalDocument doc) {
 			add(doc);
 		} else if (action == Action.DELETE) {
 			delete(doc);
+		} else if (action == Action.UPDATE) {
```
jwestberg: I think for Solr `Action.UPDATE` and `Action.ADD` should be the same. In fact, if you look at Solr, it only has UpdateRequestHandler, no AddRequestHandler, so `Action.UPDATE` should be the only supported operation, if any ;)

rgraneru: Ok. But that should be a fix for a later release. What I can do now is to
put let UPDATE and ADD both use the add(doc) method.

But the exception throwing if the action isn't set, instead of doing
nothing, isn't that bad, is it?

jwestberg: Yeah, that's what I meant, it should be `if(action == Action.ADD || action == Action.UPDATE)` I think in the case of Solr. 

Throwing an exception is fine if action is missing though, yes.

PR #122, opened 2012-11-15
Admin service delete stage

rgraneru:
That is...the pipelinewriter only sets it to inactive, but that is good enough


rgraneru:
Hmm. Was a bit too quick there. I'll clean it up and send another pull request

rgraneru:
This didn't compile after merging with the rest of hydra master

PR #123, opened 2012-11-15
Admin service delete stage

rgraneru:
Added a delete stage to the service. This is now updated with support for groups


rgraneru:
I really need to create a "TODO" when issuing a pull request. This was a lot back and forth to no gain

PR #128, opened 2012-11-22
JsoupAttrSelector

DanielGV:
Added a new stage that allows to retrieve a tag attribute from the matching elements of a Jsoup selector.

For example:
"htmlField":"extracted_html",
"jSoupConfigs":[
        {"selector":"link[rel*=canonical]",
         "fieldname":"url",
         "attribute":"href",
         "singlevalue":"true"}
]

Would retrieve the first canonical link of the form ("url" in this example):
<link rel="canonical" href="url">

### DIFF
--- a/stages/processing/web/src/main/java/com/findwise/hydra/stage/webstages/JsoupAttrSelector.java
+++ b/stages/processing/web/src/main/java/com/findwise/hydra/stage/webstages/JsoupAttrSelector.java
@@ -1,0 +1,178 @@
+package com.findwise.hydra.stage.webstages;
+
+import com.findwise.hydra.common.Logger;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+import org.jsoup.Jsoup;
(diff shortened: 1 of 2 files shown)


laserval:
I'm not sure I understand why this needs to be a separate stage. Could this functionality not be added to the JsoupSelectorStage? Optionally selecting attributes sounds like a good idea to me.

laserval:
This is better off as additional functionality in JsoupSelectorStage. Please submit a new pull request when you have implemented it in that class.

jensbengtsson:
Just feels like the JsoupSelectorStage is getting very messy, handling all kind of different configuration options. Dividing selecting tags and selecting attributes into different stages clears things up somewhat.

laserval:
Perhaps the selection code can be moved to a utility class, like in the Tika stages? Or an abstract class.

jensbengtsson:
Yeah, that might be a way. Don't know if we will have time to do the changes so we might just do our own release.

PR #131, opened 2012-11-23
tika-stages now uses shade instead of jar-with-dependencies

### LINKED ISSUE #120: Tika stage fails to extract content
Hey everyone,

I struggled last week to get TikaStage to work correctly. The problem was that Tika managed to find the correct content-type for the data, but could not extract contents. Even weirder was that I couldn't replicate the problem when running the same code from within Netbeans.

In the end, I have managed to isolate the problem to some sort of resource conflict when using "jar-with-dependencies". To fix the problem, I've instead used the onejar plugin: https://code.google.com/p/onejar-maven-plugin/

Shall I upload the change to the pom.xml? Has anyone else had the same problem?

remen:
Fixes https://github.com/Findwise/Hydra/issues/120

### DIFF
--- a/stages/processing/tika/pom.xml
+++ b/stages/processing/tika/pom.xml
@@ -58,30 +58,48 @@
 					<target>1.6</target>
 				</configuration>
 			</plugin>
-			<plugin>
-				<artifactId>maven-assembly-plugin</artifactId>
-				<configuration>
-					<finalName>${project.name}</finalName>
-					<descriptorRefs>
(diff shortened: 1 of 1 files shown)


laserval:
I can't see an easy way to solve this without using shade, so I'll merge this in to ensure we don't have the problem in the future. Few of our other stages have as big dependencies as the Tika stages, so I doubt the same problem appears there.

We'll have to look at using shade instead of assembly, and if it has any downsides (I don't really see any, except that the configuration has to list the license file extensions).

PR #134, opened 2012-11-26
Abstraction of the JSoup Selector Stages and Addition of Attribute Selector

DanielGV:


DanielGV:
Check it out, it works on out project. I have also modified some test from empty String to null matching.

PR #136, opened 2012-12-02
Implemented a cache for the DatabaseConnector

jwestberg:
This is a big one. Sorry guys!

We've implemented CachingDatabaseConnector which will use a MemoryConnector to serve documents to stages, while using MongoDB as the backing storage. 

There were _a lot_ of things changed that are not directly related to this, but became necessary: For instance, we needed to configure a Hydra Core to use this cache, and Guice became an obstacle. So we removed Guice.

Hydra is, after this merge, roughly 40% faster n our tests (a 11-stage pipeline, dropping from 30 minutes to complete to 19).


jwestberg:
Apparently, I fail at not messing up my repositories when it becomes complex :(

api/src/main/java/com/findwise/hydra/common/SerializationUtils.java
```diff
@@ -32,6 +34,42 @@
  *
  */
 public final class SerializationUtils {
+	
+	private static SimpleDateFormat getDateFormat() {
```
jwestberg, line 38: Yes, Hydra now suddenly supports dates properly. This was required for caching to work correctly

api/src/main/java/com/findwise/hydra/common/SerializationUtils.java
```diff
@@ -72,34 +109,13 @@ public static Object toObject(String json) throws JsonException {
...
-								.getTime());
-					}
-				});
-
-		gsonBuilder.setDateFormat(DateFormat.FULL);
-		return gsonBuilder.serializeNulls().create().toJson(o);
+		while(true) {
+			try {
+				return gson.toJson(o);
+			} catch(ConcurrentModificationException e) {
```
jwestberg, line 115: Serialization is a problem when it's of a document that is being kept in memory in the cache, since another thread may modify the document while we are serializing. However, we can be sure that this is a passing state, so we just retry until we're done.

core/pom.xml
```diff
@@ -141,13 +141,6 @@
 			<scope>compile</scope>
 		</dependency>
 		<dependency>
-			<groupId>com.google.inject</groupId>
-			<artifactId>guice</artifactId>
```
jwestberg, line 145: Guice be gone!

api/src/main/java/com/findwise/hydra/stage/AbstractStage.java
```diff
@@ -28,7 +28,6 @@
  */
 public abstract class AbstractStage extends Thread {
 
-	public static final String DEV_MODE_STRING = "DEV_MODE";
```
laserval, line 31: Does this mean that DEV_MODE has been replaced with the performance logging mode, or am I missing something?

jwestberg, line 31: I may have gotten a little overexcited and removed DEV_MODE when starting from the Command Line. I've wanted to do this for a while, and I guess while trying to add the PERFORMANCE_LOG parameter, I may have seized the opportunity. 

As far as I know, no one uses DEV_MODE, and it's not properly documented - using remote configuration with DEBUG is preferred (since your stage wont work without access to a Core anyway). Also, requiring JSON here in the API means we can't change serialization format without introducing jar bloat.

jwestberg:
After this merge, a side effect is that we can get CI on all things Hydra through [travis-ci](https://travis-ci.org/) :)

Issue #129, opened 2012-11-23
Stage groups endpoint in Admin service


laserval:
It should be possible to GET the currently configured Stage Groups, as well as POST stages to groups.

Currently, the stages/{stage-name} POST endpoint puts the stage in its own group.
Stage groups can be listed on the root endpoint.

jensbengtsson:
Should there be two different end points, one where you don't need a group name (aka the current one) and one where you provide a group name?

So like this

@RequestMapping(method = RequestMethod.POST, value = "/libraries/{id}/stages/{groupName}/{stageName}")

laserval:
Closed by #138

PR #138, opened 2012-12-16
Added rest endpoints for stage groups

jensbengtsson:
As suggested in https://github.com/Findwise/Hydra/issues/129 I have added end points for listing stage groups as well as adding stages to a named stage group. 

I don't know why the diffs are messed up, showing every line changed. :/

### DIFF
--- a/admin-service/README.md
+++ b/admin-service/README.md
@@ -89,7 +89,15 @@ Available endpoints
 	
 	Request content: Stage configuration in json format
 
-	Adds a stage or replaces a stage in the pipeline with the attached configuration as raw body.
+	Adds a stage or replaces a stage in the pipeline with the attached configuration as raw body. A stage group with the stage name will be created.
+	
+* **/libraries/{library-id}/stages/{stage-group}/{stage-name}** - Method: POST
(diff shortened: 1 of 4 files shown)


jensbengtsson:
https://github.com/Findwise/Hydra/pull/138/files?w=1

Show's without white space changes

laserval:
Good stuff! Could you add a short description to the README.md about the new endpoints?

jensbengtsson:
Sure, will do it later and add the commit.

laserval:
Something is messing up the formatting in README.md, could you put the same formatting around the stagegroups response as for the other stuff? I think it's three backticks.

laserval:
The reason for the diff showing every line as changed is that all Unix line endings have been replaced with Windows line endings, so every line is one character longer.

In general Hydra uses only Unix line endings, so if you encounter the same behavior from diff have a look at the line endings.

jensbengtsson:
Netbeans didn't show all lines changed when I diffed in it before commit. Neither did my git shell. So don't know why it ended up that way in the end.

PR #158, opened 2013-02-07
Fixes admin-service compilation error due to changes in DatabaseQuery

laserval:
This adds the missing `requireNotFetchedByStage()` method to AdminServiceQuery.


jwestberg:
Oops. My bad for not keeping admin-service up to date. Should we add it to the parent-pom?

laserval:
Perhaps that's the easiest solution to catch stuff like this. Tried to get Travis to do it, but it seems complicated.

jwestberg:
Ideally, it should be something like:

pom.xml
 |- application
 |        |- pom.xml
 |        |- core/
 |        |- api/ (etc)
 |- service
 |        |- pom.xml

So that we could build service and application separately, while building them together with the upper pom.xml

laserval:
Agreed!

Can you merge this fix in if it looks ok?

jwestberg:
Least complex commit ever. Merging :)

PR #165, opened 2013-02-18
Refactored Document IDs to be consistent across DatabaseConnector

jwestberg:
In order to fix various issues with caching, it became apparent that the major issue was document ids being represented as Objects. Since that was the case, we could not reliably keep one type of document id in the cache and another in the back end. 

Converting between the two became incredibly complicated, and would often completely fail.

This pull request adresses that issue. 

(Since it's a low level change on the Document API classes, it has some rather large effects, hence the 105 files changed)


database-impl/mongodb/src/main/java/com/findwise/hydra/mongodb/MongoDocumentIO.java
```diff
@@ -487,30 +497,30 @@ public boolean markFailed(DatabaseDocument<MongoType> d, String stage) {
...
+//		BasicDBObjectBuilder dbob = BasicDBObjectBuilder.start(mq.toDBObject().toMap());
+//		dbob.add(MongoDocument.METADATA_KEY+"."+Document.PENDING_METADATA_FLAG, new BasicDBObject("$exists", false));
+//		
+//		Date earlierThan = new Date(new Date().getTime()-intervalMillis);
+//		dbob.add(MongoDocument.METADATA_KEY+"."+DatabaseDocument.FETCHED_METADATA_TAG+"."+tag, new BasicDBObject("$lt", earlierThan));
+//		
+//		DBObject update = new BasicDBObject(MongoDocument.METADATA_KEY+"."+DatabaseDocument.FETCHED_METADATA_TAG+"."+tag, new Date());
+//		
+//		return findAndModify(dbob.get(), getUpdateObject(update));
+//	}
```
laserval: Why is this commented out?

database/src/main/java/com/findwise/hydra/DocumentWriter.java
```diff
@@ -51,8 +51,8 @@
 	 * should <code>intervalMillis</code> be less than the less than the time
 	 * elapsed since the document was last tagged.
 	 */
-	DatabaseDocument<T> getAndTagRecurring(DatabaseQuery<T> query, String tag,
-			int intervalMillis);
+//	DatabaseDocument<T> getAndTagRecurring(DatabaseQuery<T> query, String tag,
+//			int intervalMillis);
```
laserval, line 55: Should these methods be removed, or replaced with something else?

jwestberg, line 55: They should be removed, as they are unused, undocumented, and a pain to maintain. I fix.

database-impl/mongodb/src/main/java/com/findwise/hydra/mongodb/MongoDocumentIO.java
jwestberg: Se later note. They should be removed, as they are no longer required by the interface. I fix.

PR #173, opened 2013-03-13
Added a stage that removes html and javascript from a html page

rgraneru:
Instead of using regexp to remove html tags, this uses libraries from Jericho to remove all non-text from a html page. Even javascript.


rgraneru:
I have changed the location of the stage now. its now residing in web

PR #175, opened 2013-03-13
Changes 'examples' to depend on api instead of basic-stages.

laserval:
The dependency on basic-stages is due to old project structure where AbstractStage lived outside api.

Also adds examples to parent build.


remen:
Reviewed and tested. However, there are random test failures in other modules. Logged as #179.

PR #176, opened 2013-03-13
Deprecates input stages, fixes compilation errors for the same.

laserval:
This fixes compilation errors missed because the input stages are unused. The stages are also now deprecated and should be removed in a future release.


remen:
Reviewed and tested. _However_ stages/in/json-in depends on stages/in/server, which broke my build unless I did it in the correct order (the SNAPSHOT on our nexus was out of date). This is why I want to add all of the stages to the top level build, so I can use mavens reactor functionality to build stuff in the correct order.

Issue #133, opened 2012-11-26
Include all configurable properties in the Core resource.properties file


laserval:
The `resource.properties` file only contains some of the properties that can be set. Missing are e.g. the port number.

All properties should be present in the file - as comments or with sane defaults.

jensbengtsson:
Is there really anything missing besides the rest port number? :)

laserval:
I believe #177 added the missing property, and made it easier to understand. Closing.

PR #177, opened 2013-03-13
Adds resource.properties to bin folder alongside built hydra-core.

laserval:
Adds some comments to properties file, and the missing `rest_port` configuration option. This should be all required in #133.


remen:
Reviewed and tested.

Issue #180, opened 2013-03-16
Travis build is failing


remen:
I am guessing that this is due to that we are trying to open a port which the server does not allow:

```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running com.findwise.hydra.net.HttpRESTHandlerTest
10:24:37.691 [I/O dispatcher 1] INFO  c.f.hydra.net.HttpResponseWriter - Got ID ping!
10:24:37.743 [Thread-2] ERROR com.findwise.hydra.net.RESTServer - I/O error in RESTServer: Failure binding socket to address 0.0.0.0/0.0.0.0:20000

...

Results :
Failed tests:   testAccessRestrictions(com.findwise.hydra.net.HttpRESTHandlerTest): Server should have been working, we are connecting through localhost...
Tests run: 41, Failures: 1, Errors: 0, Skipped: 0

```

jwestberg:
Pretty sure there is something slightly more complicated afoot. There are other tests that rely on port binding, since the start method will attempt repeatedly to find an open port to use. 

I think it might be something along the lines of the Travis VM not resolving localhost to the calling IP.

remen:
Yeah, you're right. I've downloaded the travis image (vagrant ftw!) and
will debug locally. I think it is due to 127.0.0.1 resolving to
localhost.localdomain because of /etc/hosts. Will check.

Anyway, while looking at the code I have two points I would like to raise:

1) I think the logic for knownHosts is "wrong". We take the client
ip-address and do a reverse lookup to get the hostname and check if it's in
the list. I don't think this is a good idea, since a) an IP-address can be
associated with several hostnames and b) this will become strange if we add
an ip-address to the list. I think a better solution is to resolve the
hosts in the knownHosts list (i.e. not do reverse lookup), but will have to
do more testing to see whether it works as intended.

2) Why do we even have this logic? If we are going to do IP-based security,
wouldn't a firewall be better suited?

/Petter

remen:
Damn it, couldn't replicate with the vagrant image. Maybe have the wrong
version... Still digging.

jwestberg:
Petter, I think you're right on both accounts. Preferably, one might have the option of doing either option, and hostnames are probably the least important one, since it's so error-prone.

[...]

PR #181, opened 2013-03-17
Improved HttpRESTHandler allowedHosts logic (don't use reverse lookup)

remen:
This allows for ip-addresses in allowedHosts and I think it will fix the problems we're having with Travis (#180).


remen:
Yep, the travis build seems to be successful :D

jwestberg:
Good work. Will be happy to see that green color from Travis again :)

laserval:
This is great! Thanks @remen :)

PR #183, opened 2013-03-19
Use SocketAppender and slf4j as logging api

### LINKED ISSUE #108: Replacement of current API Logging facilities
StageGroups (#106) will somewhat break logging support when something is run as a group, as stages will log on the group level rather than on the stage level.

I think it could be easily replaced by a logging webservice alongside the other services in Core.

remen:
- Deprecates `com.findwise.hydra.Logger` (simply use slf4j instead)
- Uses `SocketAppender` in the stages and `SimpleSocketServer` in the core
- Removes `InternalLogger` class. Use `LoggerFactory.getLogger("internal")` instead.
- Hopefully fixes #108


remen:
Hey guys, I think I got something that is fit for review. Please have a look at the overall design and see what you think.

I am guessing that you will react to the fact that log messages are no longer shown on stdout if you run the stage manually. If that is the case, it should be easy to fix (just add a command line flag to keep the stdout appender open).

remen:
Oh, and if you are drowning in the number of files changes, I could split it into two commits where one of them only contains all of the `s/Logger/logger/g` changes :P

dsmiley:
Good!  I'm always disappointed when I see projects introduce their own loggers when SLF4J et. all work just fine.

laserval:
I'm tested this a bit and so far it looks great, will have a look at the implementation to make sure I understand it as well ;)

api/src/main/java/com/findwise/hydra/Logger.java
```diff
@@ -1,34 +1,27 @@
 package com.findwise.hydra;
 
 
+import ch.qos.logback.classic.LoggerContext;
+import ch.qos.logback.classic.net.SocketAppender;
+import org.slf4j.LoggerFactory;
```
laserval, line 6: This is the only import that is used, right? The others should be removed.

core/src/main/java/com/findwise/hydra/Main.java
```diff
@@ -29,7 +31,9 @@ public static void main(String[] args) {
 		} else {
 			conf = getConfiguration(null);
 		}
-		
+
+        new SimpleSocketServer((LoggerContext) LoggerFactory.getILoggerFactory(), conf.getLoggingPort()).start();
```
laserval, line 35: Is it ok to let this simply die when the Core is shut down, or should it have some kind of shutdown hook for stopping the SocketServer?

Also, you've used spaces instead of a tab ;)

core/src/main/java/com/findwise/hydra/CoreConfiguration.java
```diff
@@ -2,8 +2,12 @@
 
 public interface CoreConfiguration extends Configuration, DatabaseConfiguration {
 	static final String PERFORMANCE_LOGGING = "performance_logging";
-	
-	int getPollingInterval();
+	static final String LOGGING_PORT = "logging_port";
```
laserval, line 5: Please add the new configuration option to the resource.properties file. Comment it out if you want, but I think visibility is good.

core/src/main/java/com/findwise/hydra/StageRunner.java
```diff
@@ -196,14 +199,15 @@ private boolean runGroup() {
...
+                String.format("%s (stdin)", stageGroup.getName()),
+                p.getInputStream()
+            ).start();
+            new StreamLogger(
+                String.format("%s (stderr)", stageGroup.getName()),
+                p.getErrorStream()
+            ).start();
 
-            if (loggingEnabled) {
-                PumpStreamHandler streams = new PumpStreamHandler(new StreamLogger(stageGroup.getName()));
```
laserval, line 201: Since the PumpStreamHandler is no longer used, remove the import.

core/src/main/java/com/findwise/hydra/Main.java
remen, line 35: Oops :)

remen:
Excellent comments. I'll have a look at it at first opportunity (probably this weekend)

jwestberg:
Petter, I've been playing around with this, and it works great!

I have some suggested improvements that I accidentally implemented, mostly to help when debugging a stage locally, so I'll send in a pull request with that to your branch. Probably tonight.

laserval:
Will do the last fixes and submit a new pull request for this.

PR #184, opened 2013-03-20
Completely remade the caching functionality.

jwestberg:
Here's a completely new, and improved, implementation of the caching functionality. Extensively unit tested and acceptance tested, and without the issues that was plaguing the previous implementation. The code is also considerably easier to maintain the previous attempt. 

It is configurable in resource.properties whether or not to use the cache.


jwestberg:
Just noticed I had some eclipse artifacts in my poms. Will fix and open a new pull request.

Issue #157, opened 2013-02-05
DocumentCache is not removing fields


ebbesson:
Unable to remove fields in cache :(

jwestberg:
This issue is fixed with #185

PR #185, opened 2013-03-20
Completely remade the caching functionality.

jwestberg:
Here's a completely new, and improved, implementation of the caching functionality. Extensively unit tested and acceptance tested, and without the issues that was plaguing the previous implementation. The code is also considerably easier to maintain the previous attempt.

It is configurable in resource.properties whether or not to use the cache.


database-impl/mongodb/src/main/java/com/findwise/hydra/mongodb/MongoQuery.java
```diff
@@ -172,6 +228,99 @@ public String toJson() {
...
+//		List<String> list = new ArrayList<String>();
+//		
+//		for(Map.Entry<String, Object> e : lq.getContentsEquals().entrySet()) {
+//			if(e.getValue() == Boolean.FALSE) {
+//				list.add(e.getKey());
+//			}
+//		}
+//		
+//		return list;
+//	}
```
laserval: These should be removed, right?

jwestberg: Damn it. Yes. I'll fix it!

database/src/main/java/com/findwise/hydra/CachingDocumentNIO.java
```diff
@@ -3,74 +3,131 @@
...
+					d.setFetchedBy(t, new Date());
+				}
+			}
+			cache.add(list);
+		}
 
 		return list;
 	}
 
+	int i = 0;
```
laserval: What's this?

laserval:
This should resolve #157, correct?

jwestberg:
Yes. It also resolves the "unfiled" issue of the old cache being terrible ;)

Issue #198, opened 2013-04-12
Race condition when adding documents with file attachments


remen:
When adding a document with binary file attachments, there is a time-period in which the document exists in the mongo database, but the files haven't been committed yet. This leads to stages sometimes being unable to fetch the binary attachment, giving rise to log messages such as:

```
2013-04-12 11:29:21 : ERROR  : Thread-17 : ChalmersPreProcessStage : Node gave an unexpected response: HTTP/1.1 404 Not Found
2013-04-12 11:29:21 : ERROR  : Thread-17 : ChalmersPreProcessStage : Message: No file found by the name data for this document
```

rgraneru:
Is this on Unix or Windows or both?
Den 12. apr. 2013 12:20 skrev "Petter Remen" notifications@github.com
følgende:

remen:
Platform should not matter.

rgraneru:
Does should a wish or a fact? :-)  As it is now, Windows mongodb misbehaves
more than the Unix one
Den 12. apr. 2013 19:01 skrev "Petter Remen" notifications@github.com
følgende:

> Platform should not matter.
>

remen:
Hehe. Confirmed problem on both mac and linux at least. The reason for the
bug is quite simple actually, the document has to be committed to mongodb
before the binary attachments since the attachments uses the ID generated
by mongo. However, the stages happily pick up the document but fails when
trying to reference the attachments. Anyway, I have a solution strategy in
mind. Hope to send pull request next week.

PR #199, opened 2013-04-15
Added support for "atomic" insert of documents with attachments

remen:
Hey,

Here's my suggestion of a fix for the race condition when adding documents with attachments (#198). Basically, here's what's been done:

1) Added new `DocumentWriter.insert`:

``` java
boolean insert(DatabaseDocument<T> d, List<DocumentFile<T>> attachments);
```

2) Changed corresponding implementation in `MongoDocumentIO` to set a metadata flag on the document to make `getAndTag` ignore it until all the attachments have been committed.

To understand the behaviour, I suggest looking at the new tests in `MongoDocumentIOTest`.

The other commits are basically a bit of cleaning to simplify the implementation.

### DIFF
--- a/database-impl/mongodb/src/main/java/com/findwise/hydra/mongodb/MongoDocumentIO.java
+++ b/database-impl/mongodb/src/main/java/com/findwise/hydra/mongodb/MongoDocumentIO.java
@@ -255,7 +255,51 @@ public class MongoDocumentIO implements DocumentReader<MongoType>, DocumentWrite
 		}
 		return false;
 	}
-	
+
+	/* (non-Javadoc)
+	 * @see com.findwise.hydra.DocumentWriter#insert(com.findwise.hydra.DatabaseDocument)
+	 */
+	@Override
+	public boolean insert(DatabaseDocument<MongoType> d, List<DocumentFile<MongoType>> attachments) {
+		if(attachments == null || attachments.isEmpty()) {
+			return insert(d);
+		}
+
+		d.putMetadataField(MongoDocument.COMMITTING_METADATA_FLAG, true);
+
+		if (!insert(d)) {
+			return false;
+		};
+
+		if (!writeAttachments(d, attachments)) {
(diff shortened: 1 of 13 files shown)


remen:
I've changed my mind regarding the API. I'll send and updated pull request.

database-impl/mongodb/src/main/java/com/findwise/hydra/mongodb/MongoDocumentIO.java
```diff
@@ -240,20 +240,46 @@ public void deleteAll() {
...
-				documents.insert((MongoDocument) d, concern);
-				return true;
+	@SafeVarargs
+	/* Needs to be final in order for SafeVarargs to apply */
+	public final boolean insert(DatabaseDocument<MongoType> d, DocumentFile<MongoType>... attachments) {
+		if(d.getID() != null) {
+			return false;
+		}
+
+		d.putMetadataField(MongoDocument.COMMITTING_METADATA_FLAG, true);
```
jwestberg, line 268: Perhaps this should be forked off somehow at this point? We're introducing a mandatory database lock here since all documents get this flag, which then gets removed in the update method. In the `update()`-method are doing a `DBCollection.findAndModify()` by an indexed field (`_id`), so it is probably pretty cheap. But since most documents probably do not have attachments, it can be a rather expensive thing to do all the time.

remen, line 268: Good point. I'll use a separate code path when attachments is empty.

remen:
New updated pull request. Changed the API to not use varargs (too much problems with generics), and no longer performs a two-stage commit if no attachments.

remen:
Rebased against common root of 0.3.x and master to simplify merging in to both branches.

laserval:
I merged this and also cherry-picked the commits into `0.3.x`.

PR #206, opened 2013-05-24
Adds check for required parameter arguments in AbstractStage

laserval:
Stages should still implement `init()` to check validity of parameters.
Adds `required` parameter argument to relevant stage parameters.
Changes some log levels to reduce log spam.


jwestberg:
I like this and will happily merge it, but it's in conflict with #215.

 Could you rebase on master and amend this?

jwestberg:
I did some trixtering and merged it by push to upstream. Or, well, actually you did. In the past. (Yay for modifying history)

PR #207, opened 2013-05-24
MySQL Output Stage

Tethik:
MySQL Output Stage used in the crimemap project. Has not been tested against the 0.4.0 core, because the core won't build on my machine... (fails on mongodb tests)


rgraneru:
There is a known issue with unit tests failing. Use the -DskipTests option
when building core. But testing with the 0.3 release should be good enough.
Den 25. mai 2013 00:13 skrev "Joakim Uddholm" notifications@github.com
følgende:

> MySQL Output Stage used in the crimemap project. Has not been tested
> against the 0.4.0 core, because the core won't build on my machine...
> 
> ## (fails on mongodb tests)
> 
> You can merge this Pull Request by running
> 
>   git pull https://github.com/Tethik/Hydra master
> 
> Or view, comment on, or merge it at:
> 
>   https://github.com/Findwise/Hydra/pull/207
> Commit Summary
> - MysqlStage committed.
> - Kodkommentar..
> 
> File Changes
> - _A_ stages/out/mysql-out/CHANGES.txthttps://github.com/Findwise/Hydra/pull/207/files#diff-0(19)
> - _A_ stages/out/mysql-out/LICENSEhttps://github.com/Findwise/Hydra/pull/207/files#diff-1(674)
> - _A_ stages/out/mysql-out/pom.xmlhttps://github.com/Findwise/Hydra/pull/207/files#diff-2(123)
> - _A_
>   stages/out/mysql-out/src/main/java/com/minions/hydra/MySQLStage.javahttps://github.com/Findwise/Hydra/pull/207/files#diff-3(100)
> - _A_
>   stages/out/mysql-out/src/test/java/com/minions/hydra/MySQLStageTest.javahttps://github.com/Findwise/Hydra/pull/207/files#diff-4(32)
> 
> Patch Links:
> - https://github.com/Findwise/Hydra/pull/207.patch
> - https://github.com/Findwise/Hydra/pull/207.diff

stages/out/mysql-out/src/main/java/com/minions/hydra/MySQLStage.java
```diff
@@ -0,0 +1,100 @@
...
+import com.findwise.hydra.local.LocalDocument;
+import com.findwise.hydra.stage.AbstractOutputStage;
+import com.findwise.hydra.stage.Parameter;
+import com.findwise.hydra.stage.Stage;
+
+
+@Stage(description="Writes documents to a MySQL table")
+public class MySQLStage extends AbstractOutputStage {
+	
+	@Parameter(name = "columnmap")
```
laserval: You can mark the parameters that don't have any defaults as _required_ by adding `required = true` in the annotation. Since these have to be set for the stage to work, they should have defaults or be required.

stages/out/mysql-out/pom.xml
```diff
@@ -0,0 +1,123 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>com.minions.hydra</groupId>
```
laserval: The groupId should be `com.findwise.hydra`.

stages/out/mysql-out/src/main/java/com/minions/hydra/MySQLStage.java
```diff
@@ -0,0 +1,100 @@
+package com.minions.hydra;
```
laserval: The package should be `com.findwise.hydra`.

laserval:
Hi, thanks for submitting this! I'm Olof, liked what you had used this for in your CrimeMap project :)
As Roar said, the MongoDB integration tests are unstable on some machines (haven't found out why). You can run `mvn clean package` to run only the unit tests, then `mvn verify` to run the integration tests. That way you can at least get the majority of tests to run - only the MongoDB integration should have problems.

As for the stage, it looks good. Will test it a bit next week (regular job gets in the way...). Some things from looking at the code:
- Are there any good libraries that can handle the query string building?
- Not an expert on SQL, but there's no input sanitation here, right? While input in Hydra is often somewhat controlled, this is risky.
- Parameters lack defaults, you should mark them as `required` so that users know what needs to be configured.

Just to get some more info, what are you building Hydra Core with? OS, Maven version, Java version and MongoDB version?

Thanks for the contribution! It's great to see others using Hydra for new things :)

Tethik:
Hi again!

I managed to get the latest core to build, but I haven't had time to try it out with the newest build. Also the maven build file might not be correct at the moment, so I will try to fix it and the comments you mentioned. 
- Not sure if there is any good library for query building. Some googling led to: http://www.jooq.org/ and http://www.querydsl.com/. Perhaps jdbc has something too, but I think going too abstract might be bad...
- There is input sanitation against SQL injection for the value fields in the form of prepared statements. The columnmap and table name however is not sanitised, but so long as you trust your json config files it should be fine. Sanitation against for example XSS  is possible, but really that should depend on where your output data is going.
- Will do.

Hydra Core: was 0.3.2, will rebuild with 0.4.0
OS: Windows 8
Maven version: Apache Maven 3.0.4
Java version: 1.7.0_09
MongoDB version: v2.2.1

laserval:
Good points! You are right that the config should be trusted. As for query building, I think it could make sense to abstract more if there are more operations introduced (such as UPDATE and DELETE). Although they will still be pretty static.
Actually, that makes me realize another thing that I would like to see here:
- Documents in Hydra hold an _action_ that should be recognized by the output stages. All the stage has to do is check the value of the action field (`document.getAction()`) and do the appropiate action. [See the ElasticsearchOutputStage](https://github.com/Findwise/Hydra/blob/master/stages/out/elasticsearch-out/src/main/java/com/findwise/hydra/output/elasticsearch/ElasticsearchOutputStage.java#L64) for an example. Would be good to either implement this or to add a note to the stage that makes it clear only ADD is supported. As you can see, the current output stages actually do the same thing for ADDs and UPDATEs, because for search engines they usually are.

Great, thanks for the info.

laserval:
Thanks for the update, will have a look at using the stage.

Tethik:
Hmm, didn't actually mean to commit all of that to the pull request. Will have to clean this up...

laserval:
Ah, yes, please keep to the SQL stage in this pull request. The input stages we've actually deprecated.

stages/out/mysql-out/src/main/java/com/findwise/hydra/MySQLStage.java
```diff
@@ -0,0 +1,99 @@
...
+import com.findwise.hydra.stage.AbstractOutputStage;
+import com.findwise.hydra.stage.Parameter;
+import com.findwise.hydra.stage.Stage;
+
+
+@Stage(description="Writes documents to a MySQL table")
+public class MySQLStage extends AbstractOutputStage {
+	
+	@Parameter(name = "columnmap", required=true)
+	private Map<String, String> columnmap;
```
Nycander, line 19: A description of this parameter might help, it's the only one which one might wonder how to configure

stages/out/mysql-out/src/main/java/com/findwise/hydra/MySQLStage.java
```diff
@@ -0,0 +1,99 @@
...
+	private String user;
+	
+	@Parameter(name = "password", required=true)
+	private String password;
+			
+	@Override
+	public void output(LocalDocument document) {
+		if(columnmap.keySet().size() == 0)
+		{
+			Logger.error("Columnmap is empty?");
```
Nycander, line 40: If I remember correctly these kinds of checks should be performed in the init method which you can override. Correct me if I'm wrong :)

stages/out/mysql-out/src/test/java/com/findwise/hydra/MySQLStageTest.java
```diff
@@ -0,0 +1,32 @@
...
+	
+	@Before
+	public void setUp() throws Exception {
+		stage = new MySQLStage();
+	}
+
+	@Test
+	public void testCreateQuery() {		
+		String[] cols = { "name", "id", "group", "calendar" };		
+		assert(stage.createQuery(cols).equals("INSERT INTO null SET  name = ?, id = ?, group = ?, calendar = ?"));
```
Nycander, line 20: Should probably use Assert.assertEquals() here instead.

native asserts are only checked if enabled in the VM (http://thesoftwarelife.blogspot.se/2008/05/dont-forget-to-enable-java-assert.html)

stages/out/mysql-out/src/test/java/com/findwise/hydra/MySQLStageTest.java
```diff
@@ -0,0 +1,32 @@
...
+		
+		boolean gotcha = false;
```
Nycander, line 22: You should probably use some other method for checking proper exception handling here.

Here are some suggestions: https://github.com/junit-team/junit/wiki/Exception-testing
I prefer the ExpectedException rule :)

stages/out/mysql-out/src/main/java/com/findwise/hydra/MySQLStage.java
```diff
@@ -0,0 +1,99 @@
...
+		// Remove trailing ,
+		builder.deleteCharAt(builder.length()-1);
+		logger.debug(builder.toString());
+		return builder.toString();	
+	}
+	
+	public void insertToDB(String[] cols, Object[] vals) throws SQLException
+	{		
+		Connection conn = DriverManager.getConnection("jdbc:mysql://"+host+"/"+db+"?user="+user+"&password="+password);
+		PreparedStatement stmt = conn.prepareStatement(createQuery(cols));
```
Nycander, line 91: You might save some performance by doing this connection in the init method, and then only check if you need to reconnect (catch an exception?) on every document.

Tethik:
Due to time constraints I probably won't be able to fix this for a while, and if I do it'll probably be in a new fork of Hydra. I'm no git-wizard yet so I assume a new pull request will have to be created in that case.

PR #210, opened 2013-05-30
Makes all children modules use parent pom

remen:
Hey,

To simplify releases and configuration, I've made all child modules inherit from the parent pom. I've also moved common configuration (such as sourceEncoding) to the parent, since this is now inherited by the children. Moreover, the parent pom inherits from sonatype oss to get distributionManagement tags etc.

I have _not_ done any pluginManagement or dependencyManagement stuff. But at least now its possible to do so (to enforce e.g. the same version of slf4j-api in all artifacts).

Have a look and see what you think. Once this has been merged, I will try to merge master into 0.3.x and perform a 0.3.1 release.

### DIFF
--- a/admin-service/pom.xml
+++ b/admin-service/pom.xml
@@ -24,18 +24,6 @@
 	
 	<build>
 		<finalName>hydra</finalName>
-		<plugins>
-			<plugin>
-				<groupId>org.apache.maven.plugins</groupId>
-				<artifactId>maven-compiler-plugin</artifactId>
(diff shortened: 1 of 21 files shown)


remen:
Oh, and by the way, I had to move the /bin directory to /distribution/bin. The reason for this is that since all children depend on the parent, the parent project gets built _first_. But at that point, hydra-core.jar is not available. By adding a "distribution" project with a dependency on hydra-core, it is guaranteed to get built in the right order.

PR #211, opened 2013-05-30
Adds support for setting multivalued static fields.

andreassalomonsson:
This looked like a bug to me. When the same field name was present more than once in fieldNames, only the last corresponding value in fieldValues was added to the document.

### DIFF
--- a/stages/processing/basic/src/main/java/com/findwise/hydra/stage/SetStaticFieldStage.java
+++ b/stages/processing/basic/src/main/java/com/findwise/hydra/stage/SetStaticFieldStage.java
@@ -99,7 +99,7 @@ public class SetStaticFieldStage extends AbstractProcessStage {
 			valueList = new ArrayList<String>();
 		}
 		valueList.add(fieldValue);
-		doc.putContentField(fieldName, valueList.toArray());
+		doc.putContentField(fieldName, valueList.toArray(new String[valueList.size()]));
 
 	}
 
--- a/stages/processing/basic/src/test/java/com/findwise/hydra/stage/SetStaticFieldStageTest.java
+++ b/stages/processing/basic/src/test/java/com/findwise/hydra/stage/SetStaticFieldStageTest.java
@@ -1,19 +1,28 @@
 package com.findwise.hydra.stage;
 
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.junit.Assert;
 import org.junit.Test;
 
-/**
- * Adds a field with a specified value to the document.
- * 
- * @author simon.stenstrom
- */
+import com.findwise.hydra.local.LocalDocument;
+
 public class SetStaticFieldStageTest {
 
-
-
 	@Test
-	public void testOverwritePolicies() throws ProcessException {
-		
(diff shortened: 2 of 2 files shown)


stages/processing/basic/src/main/java/com/findwise/hydra/stage/SetStaticFieldStage.java
```diff
@@ -99,7 +99,7 @@ private void addValueToField(LocalDocument doc, String fieldName,
 			valueList = new ArrayList<String>();
 		}
 		valueList.add(fieldValue);
-		doc.putContentField(fieldName, valueList.toArray());
+		doc.putContentField(fieldName, valueList.toArray(new String[valueList.size()]));
```
laserval, line 102: Since this still adds the values as an array, the description of this stage should be changed to match that. It currently claims to append values to lists, which it doesn't.

laserval:
All sequence-ish fields (arrays, lists, sets) are serialized and deserialized as lists. So for any stages accessing the field after SetStaticFieldStage, it will be a List. (https://github.com/Findwise/Hydra/wiki/The-Hydra-API-package, a wiki page I should make more visible)

remen:
> All sequence-ish fields (arrays, lists, sets) are serialized and deserialized as lists.

Ahh... I see. Then I am even more convinced of our prior (private) discussion that we should constrain the LocalDocument interface to reflect this. I.e. something like:

``` java
public List<String> getStringContentField(String fieldName) {}
public List<byte[]> getBinaryContentField(String fieldName) {}
public MultiMap<String, String> getStringFields() {}
public MultiMap<String, byte[]> getBinaryFields() {}
```

remen:
> All sequence-ish fields (arrays, lists, sets) are serialized and deserialized as lists.

But then, how can this stage add a value to an existing `String[]`? Confused!

laserval:
> But then, how can this stage add a value to an existing `String[]`? Confused!

From what I can see in the API code, it can't. It will always replace the value.
https://github.com/Findwise/Hydra/blob/master/api/src/main/java/com/findwise/hydra/SerializationUtils.java

Had a closer look at the code, and it _will_ work, but only for the case where the destination field doesn't exist. Then it will be created as a `String[]` and any additional values will be added to that field. For all other cases, the adding behavior is simply ignored since the field type is never `String[]`.

I'm thinking we can quite safely replace the array logic with a list.

By the way, yes, I agree on the typed getters.

remen:
Does Andreas need to rewrite this stage for the pull request to be accepted? At least what he does is an improvement to what is already there. We could create a separate issue about this stage being a bit broken in general.

laserval:
> Does Andreas need to rewrite this stage for the pull request to be accepted? At least what he does is an improvement to what is already there. We could create a separate issue about this stage being a bit broken in general.

True. I'll make an issue and merge this.

jwestberg:
Yeah, I think that's the only reasonable thing here. The pull request is fine, but it did kind of reveal that there's an open can of worms with this stage. I've created #212 to deal with this issue (and will probably implement it while bored on a train right now)

laserval:
Doh, we both created issues. Feel free to resolve them both :)

PR #218, opened 2013-06-16
jUnit external resources

laserval:
Replaces some duplicated MongoDB connection setup and teardown in integration tests with jUnit `ExternalResource`s. Also cleans up the integration tests somewhat to remove dangling references to the MongoDB connection, which appear to be causing leftover test databases in MongoDB. Most integration tests now properly do cleanup.
Along with this, some additions had to be made to allow injecting the Mongo connection and turning of the StatusUpdater thread.

Moves the temporary directory used in the `ElasticsearchOutputStage`'s integration test to a real `tmp` directory, using jUnit `TemporaryFolder`.


laserval:
It is very easy to get unchecked warnings when working with generics, and especially when combining that with varargs ;)
I got rid of the more complex warning though, which was caused by using `Arrays.asList()` with a parameterized `DocumentFile<MongoType>`... I guess all who work with Java sooner or later end up hating type erasure.

PR #219, opened 2013-06-19
Makes it possible to configure what actions a step will process.

andreassalomonsson:
I have added a parameter to AbstractProcessStage that can be used to configure what document actions that will be processed by a stage. The default is as before, in other words all actions. I also did some refactoring of the tests so that they no longer uses multiple threads.

### DIFF
--- a/api/src/main/java/com/findwise/hydra/stage/AbstractProcessStage.java
+++ b/api/src/main/java/com/findwise/hydra/stage/AbstractProcessStage.java
@@ -1,15 +1,19 @@
 package com.findwise.hydra.stage;
 
 import java.io.IOException;
+import java.util.EnumSet;
+import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import org.apache.http.ParseException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import com.findwise.hydra.Document.Action;
 import com.findwise.hydra.JsonException;
 import com.findwise.hydra.local.LocalDocument;
 import com.findwise.hydra.local.RemotePipeline;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * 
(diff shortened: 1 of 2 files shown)


laserval:
Note that this will still make the document `touched`, right?

laserval:
I have some musings about this:

The problem this pull request works around is when you have a single output stage and multiple stages that run fairly sequentially before that - and then you want to send some document directly to the outputstage. If you use `touched` queries to ensure stage order, this kind of workaround is needed to allow stages to touch documents/events that don't need processing but still need to get to that outputstage. Ideally, we'd have OR in the query. As it is, `exists` or `equals` can be used along with a signaling field or similar to get the outputstage to accept documents that haven't gone through all stages. But that requires yet another stage for doing the signal field logic. It would be cleaner to only need stages for processing and move pipeline logic to the queries.

I was also under the impression there's no `action` part in a stage query, but there is - https://github.com/Findwise/Hydra/blob/master/api/src/main/java/com/findwise/hydra/local/LocalQuery.java#L96
This combined with the pull request is a bit weird. 

What are other approaches for doing this?

remen:
I think I would like a more concrete use case to see how this fits into the
big picture. Could you, Andreas, tell us how the pipeline looks and what
you are trying to do?

andreassalomonsson:
I don't want to do any processing for deletes. The only stage that does something useful with deletes is the output stage in my case. So for adds, in the output stage, I want a document to be touched by some stages. For deletes there are no stages that are required to be touched. For all actions (add, update and delete) I also want to have a certain value in a field.

It would be nice if this could be done in a query, but I don't think it is at the moment. My workaround at the moment is to start a lot of the process methods with an if statement that checks the action and does nothing for deletes. That's why i thought it would be nice if it could be configured. But it would be much better if this could be done with a query.

remen:
My very opinionated opinion follows:

I feel that there is a strong danger in adding these "checkbox" like features. This is in my opinion an antipattern (that also exists in openpipeline) that comes from the fact that "stages" and "pipelines" are abstractions that fail to give you basic programmatic constructs such as if-statements.

In fact, when using hydra recently, I noticed that I was more and more ending up with custom-written "mega-stages" that performed all parts of the processing, since this allowed me to use all the programming constructs that I was used to (classes & methods, composition, dependency injection, _integration testing_ !!!).

So, unfortunately my not so helpful input to your problem is simply the following: Unless you _seriously_ need the customer to be able to reconfigure everything using the runtime json configuration, simply wrap the entire processing pipeline, except possibly the output stage, in a single processing stage. _That will give you all the freedom you need_. If you really need them, you can instantiate and call existing stages from within Java (however, the initialization is a bit tricky). However, in my experience, it's easier to simply look at an existing stage and reimplement exactly the functionality that you need. 

Also, if you do this, then you can start writing proper tests that tests the behavior of the entire pipeline. Believe me, you'll feel a whole lot better.

laserval:
The query should be able to provide most of the logic that is needed, I'd say. It makes sure stages only process the documents that they need, and only in the order that is necessary. Currently, though, the query is limited to forking the pipeline, and there's no convenient way of merging the pipeline again if you want to limit the query to checking `touched` state. With some logical operators in the query, this could be done.

But yes, writing mega-stages give all the things that the languages already has. The only thing you really lose is the sandboxing of processing in different JVMs - but the cases where that is needed are few anyway.

remen:
If we are going to go the path of strengthening the query stuff, I would even suggest thinking about how we could separate the concern "what position does a stage have in the pipeline" from the concern "what settings does a stage have for doing processing".

In other words, message routing could be configured seperately. This might help the implementation and also allow us to think about doing routing differently than the current "query / poll-based" approach (with it's inherent limitations wrt to performance).

laserval:
As per the discussion, it seems to me this shouldn't be merged. It duplicates query functionality and doesn't solve the underlying problem with the query/pipeline design in Hydra.

Closing it.

PR #224, opened 2013-07-02
Admin service delete documents

rgraneru:
Added a admin-service for discarding documents. Can delete one or many documents based on query. Default limits to 1 and also supports skip parameter

### DIFF
--- a/admin-service/src/main/java/com/findwise/hydra/admin/rest/ConfigurationController.java
+++ b/admin-service/src/main/java/com/findwise/hydra/admin/rest/ConfigurationController.java
@@ -174,7 +174,7 @@ public class ConfigurationController {
 	@RequestMapping(method = RequestMethod.GET, value = "/documents/discard")
 	public Map<String, Object> discardDocuments(
 			@RequestParam(required = true, value = "q") String jsonQuery,
-			@RequestParam(required = false, defaultValue = "1", value = "limit") int limit,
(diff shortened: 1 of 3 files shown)


admin-service/src/main/java/com/findwise/hydra/admin/rest/ConfigurationController.java
```diff
@@ -169,6 +169,15 @@ public StageGroup getStageGroup(@PathVariable(value = "stageGroup") String stage
 			@RequestBody String changes) {
 		return documentService.updateDocuments(jsonQuery, limit, changes);
 	}
+
+        @ResponseBody
+	@RequestMapping(method = RequestMethod.GET, value = "/documents/discard")
+	public Map<String, Object> discardDocuments(
+			@RequestParam(required = true, value = "q") String jsonQuery,
+			@RequestParam(required = false, defaultValue = "1", value = "limit") int limit,
```
laserval: Does this default limit mean that only one document gets discarded when only the query is supplied? I would expect all documents by default.

rgraneru: Yes. Default is just one document. I thought that deleting is scary so I
wanted to protect the user.

But I can agree that this is too wimpy. I can change the default to all of
you want to.
Den 2. juli 2013 09:14 skrev "Olof Nilsson" notifications@github.com
følgende:

Issue #208, opened 2013-05-29
Admin Service throws NPE when authentication fails


laserval:
When the Admin Service fails authentication towards MongoDB (e.g. when there is no credentials configured), all requests return a NPE.

laserval:
Fixed in #225

PR #225, opened 2013-07-04
Admin-service configuration via property file

laserval:
The admin-service can now be configured with a property file, just like the Core.

This means the admin-service now supports authentication to MongoDB.

Adds a stub of a test for the service endpoint methods, suggestions on how to create proper tests for those are very welcome!

### DIFF
--- a/admin-service/src/main/java/com/findwise/hydra/admin/ConfigurationService.java
+++ b/admin-service/src/main/java/com/findwise/hydra/admin/ConfigurationService.java
@@ -5,6 +5,7 @@ import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Map;
 
+import com.findwise.hydra.DatabaseException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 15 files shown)


laserval:
This fixes #208.

admin-service/src/main/java/com/findwise/hydra/admin/ConfigurationService.java
```diff
@@ -31,33 +26,37 @@ public ConfigurationService(DatabaseConnector<T> connector) {
 		this.connector = connector;
 
 		try {
-			connect();
-			pipelineScanner = new PipelineScanner<T>(connector.getPipelineReader());
+			initialize();
```
Nycander, line 30: Why perform this extraction? You had a change to have pipelineScanner final :(

admin-service/src/main/java/com/findwise/hydra/admin/ConfigurationService.java
```diff
@@ -65,39 +64,22 @@ public void addLibrary(String id, String filename, InputStream stream) {
...
-				}
-			}
-		} catch (IOException e) {
-			logger.error("Unable to get stage classes", e);
-		}
-		return map;
-	}
 
-	public Map<String, Object> getStats() {
+	public Map<String, Object> getStats() throws IOException {
```
Nycander: What could a caller possibly do with an IOException? Wrap it into something more useful perhaps?

admin-service/src/main/java/com/findwise/hydra/admin/ConfigurationService.java
```diff
@@ -31,33 +26,37 @@ public ConfigurationService(DatabaseConnector<T> connector) {
...
 	}
 
-	public void addLibrary(String id, String filename, InputStream stream) {
-		connector.getPipelineWriter().save(id, filename, stream);
+	public void addLibrary(String id, String filename, InputStream stream) throws IOException {
+		getConnector().getPipelineWriter().save(id, filename, stream);
 	}
 
-	public Map<String, Object> getLibraries() {
+	public Map<String, Object> getLibraries() throws IOException {
```
Nycander: As a user of this class, it would feel very weird to catch an IOException when looking at the rest of the signature of this method.

I think you should  wrap it into something a caller might understand and do something about.

admin-service/src/main/java/com/findwise/hydra/admin/JarFetcher.java
```diff
@@ -0,0 +1,66 @@
...
+		this("hydra-admin-tmp", ".jar");
+	}
+
+	public JarFetcher(String prefix, String suffix) {
+		this.tempPrefix = prefix;
+		this.tempSuffix = suffix;
+	}
+
+	/**
+	 * TODO: Avoid creating new files for each call to this method
```
Nycander, line 36: Is there an issue for this?

admin-service/src/main/java/com/findwise/hydra/admin/ConfigurationService.java
laserval, line 30: The pipelineScanner needs to be replaced on connection loss.

There are probably more things that should change in this service, and the connector API.

admin-service/src/main/java/com/findwise/hydra/admin/JarFetcher.java
laserval, line 36: There is now! #244

Nycander, line 36: Yay! :smiley:

laserval:
I'll merge this after having it running for quite a while in a production environment. There might be some more fixes coming in as well.

PR #228, opened 2013-07-16
Getting started simplified

ssimon:
Tried to make it easier to get started with hydra, by making use of the debugging package.

### DIFF
--- a/stages/debugging/debugging/pom.xml
+++ b/stages/debugging/debugging/pom.xml
@@ -3,19 +3,24 @@
 	<parent>
 		<groupId>com.findwise.hydra</groupId>
 		<artifactId>stages</artifactId>
-		<version>0.4.0-SNAPSHOT</version>
+		<version>0.3.1</version>
 		<relativePath>../..</relativePath>
 	</parent>
-	<artifactId>hydra-debugging-stage</artifactId>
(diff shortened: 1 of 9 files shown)


stages/debugging/debugging/pom.xml
```diff
@@ -3,19 +3,24 @@
 	<parent>
 		<groupId>com.findwise.hydra</groupId>
 		<artifactId>stages</artifactId>
-		<version>0.4.0-SNAPSHOT</version>
+		<version>0.3.1</version>
```
laserval: Is this right?

ssimon: Hmm.. I guess not.. I think I did that changes for 0.3.1 but it should work
for 0.4.0-SNAPSHOT aswell..

laserval: Ah, I see, you were working with the 0.3.x branch. Right now there aren't any breaking changes on master I believe, so it should be fine.

But yeah, if it's merging to master and isn't a tagged release, it should be a SNAPSHOT. Btw you set the test-suite artifact to 0.4.0-SNAPSHOT so...

ssimon: Fixed!

Issue #168, opened 2013-03-06
Admin service returns OK and "Added" for updated stage configs without updating them


laserval:
POSTing a stage configuration with the admin-service, then changing it and POSTing again will cause a stage restart but no configuration update.

laserval:
I am having a look at this and have found that `StageGroup` has some issues with adding/removing stages, due to using a `Set` of stages that are identified by their `hashCode`, which happens to depend on their modification date. If for some reason a stage config is added twice to a group without the group being written to the database, the order the configurations are applied is not guaranteed.

PR #231, opened 2013-08-08
Fixes one cause of #168

laserval:
`StageGroup`s were keeping the stage configurations in a HashSet, where the hashcode for each configuration was dependent on when it was created. This means a stage could get several configurations in the same stage group. When the set of stages was iterated on, the order was not guaranteed which often meant the old config was applied instead of the new.

The set is now a map using the stage name as key instead.

This should solve part of #168, although I believe there is more going on that causes the unreliable config updates.

### DIFF
--- a/core/src/main/java/com/findwise/hydra/StageRunner.java
+++ b/core/src/main/java/com/findwise/hydra/StageRunner.java
@@ -99,7 +99,7 @@ public class StageRunner extends Thread {
 		stageDestroyer = new StageDestroyer();
 
 		setParameters(stageGroup.toPropertiesMap());
-		if (stageGroup.getStages().size() == 1) {
+		if (stageGroup.getSize() == 1) {
 			//If there is only a single stage in this group, it's configuration takes precedent
 			setParameters(stageGroup.getStages().iterator().next().getProperties());
 		}
@@ -133,7 +133,7 @@ public class StageRunner extends Thread {
 			return;
 		}
 
-		if (stageGroup.getStages().size() < 1) {
+		if (stageGroup.getSize() < 1) {
 			logger.info("Stage group " + stageGroup.getName() + " has no stages, and can not be started.");
 			return;
 		}
--- a/core/src/main/java/com/findwise/hydra/StageRunner.java
+++ b/core/src/main/java/com/findwise/hydra/StageRunner.java
@@ -133,7 +133,7 @@ public class StageRunner extends Thread {
 			return;
 		}
 
-		if (stageGroup.getSize() < 1) {
+		if (stageGroup.isEmpty()) {
 			logger.info("Stage group " + stageGroup.getName() + " has no stages, and can not be started.");
 			return;
 		}
--- a/database/src/main/java/com/findwise/hydra/StageGroup.java
+++ b/database/src/main/java/com/findwise/hydra/StageGroup.java
@@ -5,6 +5,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import java.util.Collections;
 
 import org.slf4j.LoggerFactory;
 import org.slf4j.Logger;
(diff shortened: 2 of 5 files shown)


database/src/main/java/com/findwise/hydra/StageGroup.java
```diff
@@ -154,15 +154,11 @@ else if(!files.containsKey(s.getDatabaseFile().getId())) {
 	}
 	
 	public Set<String> getStageNames() {
-		HashSet<String> names = new HashSet<String>();
-		for(Stage stage : stages) {
-			names.add(stage.getName());
-		}
-		return names;
+		return stages.keySet();
```
Nycander: Should not this method create a copy of the set just as `getStages()`?

remen: I really like guava's immutable collections for circumstances like these.

Nycander: Yeees, but if we don't want to add that dependency we could just do:

``` java
Collections.unmodifiableSet(new HashSet<>(stages.keySet()));
```

Should provide the same effect, right?

remen: Well, not exactly: http://stackoverflow.com/a/5611347/1525691

Why are we afraid of dependencies?

Nycander: Yes, it won't have the same effect if someone steals the reference to the newly created hashset.
In this case, that chance is extremely low.

I'm not afraid of dependencies, I just think that they should be treated with caution.
Is it really worth it to add 2.1MB to your artifact (and to any artifact which uses your artifact as a dependency) just so you can use one utility method? I think Google Go has some wise philosophies in this regard.

laserval:
Some manual testing has shown this doesn't solve all of #168 :(

We'll have to test both the CmdLineInserter, the database packages and the Admin service to see if the problem is shared among them or if it's specific to the Admin Service.

database/src/main/java/com/findwise/hydra/StageGroup.java
```diff
@@ -37,7 +38,7 @@
 	private static final Logger logger = LoggerFactory.getLogger(StageGroup.class);
 	
 	public StageGroup(String name) {
-		stages = new HashSet<Stage>();
+		stages = new HashMap<String, Stage>();
```
Nycander, line 41: Is thread-safety a concern in this class? if yes, perhaps ConcurrentHashMap should be considered?

database/src/main/java/com/findwise/hydra/StageGroup.java
```diff
@@ -22,7 +23,7 @@
 	public static final String CLASSPATH_KEY = "classpath";
 	public static final String JAVA_LOCATION_KEY = "java_location";
 	
-	private Set<Stage> stages;
+	private Map<String, Stage> stages;
```
Nycander: final?

core/src/main/java/com/findwise/hydra/StageRunner.java
```diff
@@ -120,7 +120,7 @@ public void run() {
     		return;
     	}
     	
-    	if(stageGroup.getStages().size()<1) {
+    	if(stageGroup.getSize() < 1) {
```
Nycander: Is this performed often? Then perhaps another method is needed?

``` java
if (stageGroup.isEmpty())
```

database/src/main/java/com/findwise/hydra/StageGroup.java
laserval, line 41: It shouldn't be. Can't see any place where the group object is shared.

laserval:
Rebased on master and added an isEmpty method, as well as some logging for debugging purposes.

laserval:
Merged after having run extensively. Seems the updates are going through without failures now, although I have mostly used the admin service.

PR #233, opened 2013-08-14
SimpleFetchingTikaStage should close its resources after fetching

Nycander:
There is an issue with fetching from windows file urls where the SimpleFetchingTikaStage leaves behind open and therefor locked files.

This patch resolves those issues and well as cleans up some code.


remen:
Nice work; merging. Preferably, there would have been a test case showing that the previous code was faulty, but it seems too tricky to bother doing.

PR #242, opened 2013-08-29
Changes port for SimpleFetchingTikaStage wiremock to be a less common one

laserval:
The WireMock in SimpleFetchingTikaStageTest used the default port of 8080. This fails the test when run in environments that have a Tomcat running, such as build servers running Jenkins.

Chose a high port number that probably isn't as used. It would probably be better to choose a random open port, but WireMock seems to lack that feature. Perhaps a Rule can be made that does the check, extending the WireMockRule.

This is at least a temporary solution to get the test to run.


laserval:
I merged this myself to allow builds on CI servers running in Tomcat. It really should be a randomly assigned open port though...

Issue #249, opened 2013-09-12
Uncaught exceptions is not shutting down hydra


ssimon:

laserval:
Closed by #264

Issue #248, opened 2013-09-12
Hydra process is not shut down when it crashes


ssimon:
The SimpleSocketServer never terminates

laserval:
Closed by #264

PR #250, opened 2013-09-12
Fixes for #247, #248 and #249

ssimon:


api/src/main/java/com/findwise/hydra/stage/AbstractProcessStage.java
```diff
@@ -111,7 +114,30 @@ public void run() {
 					try {
 						logger.debug("Got new doc " + doc.getID()
 								+ " to process.");
-						process(doc);
+						ProcessThread processThread = new ProcessThread();
+						processThread.setDocument(doc);
```
Nycander, line 128: Perhaps the thread could be given a user friendly name as well?

api/src/main/java/com/findwise/hydra/stage/AbstractProcessStage.java
```diff
@@ -136,4 +162,35 @@ public void run() {
 			}
 		}
 	}
+	
+	class ProcessThread extends Thread {
+		private LocalDocument doc;
+		private Exception exception = null;
+
+		private boolean completed = false;
```
Nycander: Should be declared volatile (http://stackoverflow.com/questions/106591/do-you-ever-use-the-volatile-keyword-in-java)

I could recommend the book "Java concurrency in practice" if you have not read it :)

api/src/main/java/com/findwise/hydra/stage/AbstractProcessStage.java
```diff
@@ -136,4 +162,35 @@ public void run() {
...
+			return exception;
+		}
+
+		@Override
+		public void run() {
+			try {
+				logger.trace("Processing document");
+				process(doc);
+			} catch (Exception e) {
+				exception = e;
```
Nycander, line 207: no logging here? seems dangerous to depend on the user of ProcessThread to correctly check and call getException

api/src/test/java/com/findwise/hydra/stage/AbstractProcessStageTest.java
```diff
@@ -270,6 +270,44 @@ public void process(LocalDocument doc) throws ProcessException {
...
+		ws.setParameters(map);
+
+		LocalDocument ld = mock(LocalDocument.class);
+
+		when(rp.getDocument(any(LocalQuery.class))).thenReturn(ld);
+		when(rp.releaseLastDocument()).thenReturn(true);
+		
+		ws.start();
+		
+		Thread.sleep(1100);
```
Nycander: Depending on actual system time in tests is a bad thing (TM).

May I suggest that you create an interface which provides time? 

``` java
public interface TimeProvider {
   long currentTimeMillis();
}

public enum SystemTimeProvider implements TimeProvider {
   INSTANCE;

   @Override public long currentTimeMillis() {
         return System.currentTimeMillis();
    }
}
```

Using this you can dependency inject time, giving you the power to have tests which runs in < 1ms.
This is important.

core/src/main/java/com/findwise/hydra/Main.java
```diff
@@ -73,30 +97,86 @@ public static void main(String[] args) {
...
 			logger.error("Unable to start nodemaster... Shutting down.");
+			shutdown();
+		}
+	}
+
+	private static void shutdown() {
+		logger.info("Got shutdown request...");
+		shuttingDown = true;
+
+		long killDelay = 30 * 1000;
```
Nycander:
``` java
TimeUnit.SECONDS(30).toMillis()
```

is more readable

core/src/main/java/com/findwise/hydra/StageRunner.java
```diff
@@ -137,7 +137,7 @@ public void run() {
                 logger.error("The stage group " + stageGroup.getName() + " did not start. It will not be restarted until configuration changes.");
                 return;
             }
-        } while (timesToRetry == -1 || timesToRetry >= timesStarted);
+        } while ((timesToRetry == -1 || timesToRetry >= timesStarted) && !Main.isShuttingDown());
```
Nycander: This static referencing is extremely scary and is hard to test.

May I suggest extracting an interface which contains the method isShuttingDown and dependency inject that?

ssimon:
Great comments, I'll have a look at it later this week.

core/src/main/java/com/findwise/hydra/StageRunner.java
ssimon: Not sure I follow what you mean here. An interface that the Main class implements? Or that the StageRunner implements? And from where would I dependency inject that?

Nycander: That the main class implements.

like:

``` java
public interface TerminationMonitor {
    boolean isShuttingDown(); // or isTerminating? :o
}
```

Let Main implement that and let StageRunner get a reference to Main when it is constructed.

ta-da - testable! :grinning:

stages/debugging/debugging/src/main/java/com/findwise/hydra/debugging/InfinitLoopStage.java
```diff
@@ -0,0 +1,27 @@
...
+import com.findwise.hydra.stage.Stage;
+
+@Stage(description="A stage that gets stuck in an infinit loop and never returns")
+public class InfinitLoopStage extends AbstractProcessStage {
+
+	Logger logger = LoggerFactory.getLogger(InfinitLoopStage.class);
+	
+	@Override
+	public void process(LocalDocument doc) throws ProcessException {
+		for (int i=0;true;i++) {
```
Nycander: Whitespacing here is weird, my OCD is missing spaces after semicolon.

api/src/main/java/com/findwise/hydra/stage/AbstractProcessStage.java
ssimon: Great, I really should.. Haven't worked with java concurrency enough... Also, I'll take my own note to another commit, and change this to being "running" instead...

api/src/main/java/com/findwise/hydra/stage/AbstractProcessStage.java
ssimon, line 207: I'll add a trace logging, but depending on the actual exception you may want to ignore it in your use case. Logging it here and then again where you use it doesn't make sense.

Nycander, line 207: Perhaps what makes more sense is if you can register a "ProcessThreadExceptionListener" of sorts, which will be called upon when ProcessThread encounters an exception. 

That way the exception will get lost if and only if you as a caller chooses to ignore it.

ssimon:
Okay, everything updated. We are now instantiating Main in the main method and removing all static modifiers from that class

api/src/main/java/com/findwise/hydra/stage/AbstractProcessStage.java
```diff
@@ -111,7 +124,33 @@ public void run() {
 					try {
 						logger.debug("Got new doc " + doc.getID()
 								+ " to process.");
-						process(doc);
+						ProcessThread processThread = new ProcessThread(getStageName());
```
remen, line 127: Have you considered using futures and executors for this instead of manually creating threads?

api/src/main/java/com/findwise/hydra/stage/AbstractProcessStage.java
```diff
@@ -111,7 +124,33 @@ public void run() {
...
+						processThread.start();
+
+						while (processThread.isRunning()) {
+							Thread.yield();
+							if (timeout > 0 && timeProvider.getCurrentTime() - startTime > timeout) {
+								String docId = null;
+								if (doc.getID() != null) {
+									docId = doc.getID().toString();
+								}
+								throw new ProcessException(
```
remen, line 141: I'm not so sure about this. Let's say the stage fails to process the document within the time limit. Then we throw a ProcessException, but don't stop the processThread. What is to say that the processThread will stop at all? Won't this eventually lead to thread starvation?

api/src/main/java/com/findwise/hydra/stage/AbstractProcessStage.java
```diff
@@ -111,7 +124,33 @@ public void run() {
...
-						process(doc);
+						ProcessThread processThread = new ProcessThread(getStageName());
+						processThread.setDocument(doc);
+						
+						long startTime = timeProvider.getCurrentTime();
+						processThread.setDaemon(true);
+						processThread.start();
+
+						while (processThread.isRunning()) {
+							Thread.yield();
```
remen, line 135: I would consider doing Thread.sleep(100ms) or similar to minimize time spent in this loop. Or even better, by using futures, you could do future.get(timeout, TimeOut.MILLISECONDS) to avoid having the loop at all!

api/src/main/java/com/findwise/hydra/stage/AbstractProcessStage.java
```diff
@@ -111,7 +124,33 @@ public void run() {
...
+								throw new ProcessException(
+										"Processing of " + docId + " by " + getStageName() + " was not done within configured timout");
+							} else if (logger.isTraceEnabled()) {
+								logger.trace("Waiting for processing. Waited "
+										+ (timeProvider.getCurrentTime() - startTime)
+										+ "ms so far...");
+							}
+						}
+
+						if (processThread.getException() != null) {
```
remen, line 150: By using futures, you would get this type of functionality "for free":

``` java

Future<Object> future = executorService.submit(new Callable<Object>() {
  public Object run() throws Exception {
    stage.process(doc);
    return null; // We need Callable to be able to throw checked exceptions, but then we need to return something.
  }
}

try {
    future.get(timeout, TimeOut.MILLISECONDS);
} catch ( TimeoutException e ) {
    // Decide what to do here. Throwing ProcessException might not be the correct solution considering comment above.
} catch ( ExecutionException e ) {
    throw e.getCause(); // Throw the exception thrown by stage.process(doc)
}
```

This code is from memory, so may not work "as-is", but the pattern is valid. To get an executorService, I would consider simply doing `Executors.newFixedThreadPool(1)`, in other words, there is a single thread available for processing. Otherwise, you will need to consider thread safety, or create a Stage instance per future.

remen:
`mvn test` is failing on `hydra-api`

```
$ mvn test -Dsurefire.useFile=false
...
testProcessTimeout(com.findwise.hydra.stage.AbstractProcessStageTest)  Time elapsed: 0.014 sec  <<< FAILURE!
Wanted but not invoked:
localDocument.addError(null, <any>);
-> at com.findwise.hydra.stage.AbstractProcessStageTest.testProcessTimeout(AbstractProcessStageTest.java:300)
Actually, there were zero interactions with this mock.
...
```

laserval:
Sorry for the slow update on this. This is a great contribution, but I think we can do some things to improve it further.
As @remen commented, Futures eliminates the need for some of this home-grown functionality. There is also the question of what happens when a library repeatedly fails to return and the thread refuses to listen to interrupts - will this keep spawning new threads to execute? Ideally, when the stage continuously fails to do something it would signal to the Core that this is the case and allow Core to force a restart.

There is also the case of StageGroups, and how this behaves inside those. I'll fork this pull request and have a look at this.

laserval:
I've made a first pass of this, using `Future`s and a fixed-size threadpool. If you set a timeout, any timeouts where the processing thread can't be interrupted will cause JVM shutdown.

See:
https://github.com/laserval/Hydra/commit/d0465348114cd84f82a66ff7ead3c5dc24bdcd9a

laserval:
I believe I have a good-enough version of this in #264 , which includes the commits in this PR. Closing this.

PR #252, opened 2013-09-17
SetStaticFieldStage no longer checks for the mere existance of a field, ...

Nycander:
...but looks at the value to see if it is empty as well

### DIFF
--- a/stages/processing/basic/src/main/java/com/findwise/hydra/stage/SetStaticFieldStage.java
+++ b/stages/processing/basic/src/main/java/com/findwise/hydra/stage/SetStaticFieldStage.java
@@ -1,6 +1,7 @@
 package com.findwise.hydra.stage;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


stages/processing/basic/src/main/java/com/findwise/hydra/stage/SetStaticFieldStage.java
```diff
@@ -8,35 +9,52 @@
...
+				addValueToField(doc, key, entry.getValue());
 			} else if (overwritePolicy == Policy.THROW) {
-				throw new ProcessException("Field " + entry.getKey()
+				throw new ProcessException("Field " + key
 						+ " already has a value!");
 			}
 		}
 	}
 
+	private boolean missingContent(LocalDocument doc, String key) {
```
ssimon: I think that it would be more clear to have a hasContent function. hasContent can be used for both has content and doesnt have content, while missingContent doesnt make sense to use to find out if it has content.

Nycander: Someone on the internet agrees with you: http://www.cwu.edu/~gellenbe/javastyle/method.html

I fix.

Nycander: Fixed now.

stages/processing/basic/src/test/java/com/findwise/hydra/stage/SetStaticFieldStageTest.java
```diff
@@ -5,11 +5,11 @@
 import java.util.HashMap;
 import java.util.List;
 
-import junit.framework.Assert;
-
 import org.junit.Before;
 import org.junit.Test;
 
+import junit.framework.Assert;
```
laserval, line 11: Should be `org.junit.Assert`

PR #253, opened 2013-09-17
Fixes failing test when you are on a network which recognizes unknownhost as a valid hostname

Nycander:


Nycander:
You should be able to test this by adding unknownhost to your host-file.

PR #254, opened 2013-09-17
Adds port to configuration.

### LINKED ISSUE #251: Add database port number to configuration
Currently the default port is used for the database, and no configuration option is visible for changing the port.

Nycander:
Attempts to close #251.

I'm sorry for the whitespace changes :disappointed:


laserval:
It should be better to use a `MongoClientURI` and thus be able to support both standalone MongoDB and replica sets. See #256

Closing this, since introducing the change and then replacing it seems wasteful.

laserval:
#258 replaces this.

PR #257, opened 2013-09-20
Fixes to stage parameter parsing

laserval:
- Fixes missing check for required parameters. All stages will now once again fail to start if a `required`parameter is missing.
- Fixes problems when giving parameters a different name than their Java field name. Any `name` given in the `@Parameter` annotation will take precedence over the member variable name.
- Changes the annotated name of the configuration map in `SetStaticFieldStage` to match the name of the variable (as that was previously the correct name)


laserval:
Merged after confirmed working in production environment.

Issue #256, opened 2013-09-19
Support MongoDB replica sets


laserval:
The current MongoDB database implementation specifies a single instance. When using the Java MongoDB driver, in order to connect to a replica set or sharded cluster, the Mongo connection _must_ be instantiated with a list of hosts.

The current MongoDB database implementation does not support replica sets.

note: Practically, anyway. It should be possible to connect to the primary replica, but since that can change at any time it is hardly a useful way to connect.

laserval:
I've experimented a bit with this, and by using `MongoClientURI` and specifying the database url as `mongodb://host:port,host2,port2` it's easy to support both standalone Mongo instances and replica sets.

laserval:
Fixed by #258

PR #258, opened 2013-10-01
MongoDB client update and database configuration fixes

### LINKED ISSUE #255: Replace uses of Mongo with MongoClient
The `Mongo` class has been superseded by `MongoClient`, and will be deprecated in the MongoDB Java driver.

Note: Default `WriteConcern` has changed in `MongoClient`. This should only matter in places where we do not specify a `WriteConcern`.

laserval:
_NOTE:_ This PR makes a first pass over the configuration classes and tries to separate database and Core configuration somewhat. As such, some configuration parameters have **changed**. Notably, `core.pipeline` is now `database.pipeline`. Additionally, default values **may have changed** due to defaults declared in several places being replaced with a single declaration.
Please make sure to keep track of your resource.properties current values when testing this. Default values may differ from your own, which could cause problems.

This fixes #255 and #256 by introducing the `MongoClientURI` format for specifying the MongoDB connection: `mongodb://host1:port1,host2:port2`.
Hydra can be connected to a standalone MongoDB by simply using a single host. To connect to a replica set, specify several hosts.

All information that is required by a database implementation is now in `DatabaseConfiguration` , and in `MongoConfiguration`. All default values are bound to the specific implementation instead of being in Core - so the values make sense for the implementation. I've replaced any home-grown configurations with `MongoConfiguration`. That class should also be usable by any systems using the database packages for communicating with Hydra.

I think there is much more that can be done here, so suggestions are very welcome.


core/src/main/java/com/findwise/hydra/FileConfiguration.java
```diff
@@ -1,5 +1,6 @@
 package com.findwise.hydra;
 
+import com.findwise.hydra.mongodb.MongoConfiguration;
```
laserval, line 3: This is silly, but I'm not sure what to do to avoid this without doing more extensive refactoring of the configuration. Following the "convert" pattern in Hydra, there should be a generified configuration perhaps.

remen:
This pull request is becoming stale. What should we do with it. Do you still want it merged @laserval?

laserval:
I do, but I can take care of updating it to the current master.

remen:
Ok, good. Add a comment to the pull request once you've updated (modifying the pull request does not send out notifications it seems).

laserval:
Updated now!

remen:
Looks good (with some non-action comments). Merging.

laserval:
Yeah, I was mostly trying to remove the need for creating subclasses to DatabaseConfiguration alla over the place (one in admin service, one in examples, one in core, one in every integrator that uses database etc).

I think it needs to be replaced completely.

Issue #259, opened 2013-10-01
Complication errors for database/src/main/java/com/findwise/hydra/DatabaseException.java. Class overrides explicit Java7 constructor


ebbesson:
Constructor :
Exception(String message, Throwable cause, boolean enableSuppression, boolean writableStackTrace) does not exist in Java 6 and this causes a compliation error when trying to complie Hydra with Java 6

See
Java 6 :http://docs.oracle.com/javase/6/docs/api/java/lang/Exception.html
Java 7: http://docs.oracle.com/javase/7/docs/api/java/lang/Exception.html

laserval:
Good catch.

laserval:
Added java 6 build to the Travis CI conf on master.

PR #260, opened 2013-10-01
Adds Jsonp support to Admin Service. Jetty:run build option to admin service as well as resolves issue #259

ebbesson:


admin-service/src/main/java/com/findwise/hydra/admin/rest/JsonpCallbackFilter.java
```diff
@@ -0,0 +1,321 @@
...
+            getResponse().setContentType(getContentType());
+            getResponse().getOutputStream().write(bytes);
+            getResponse().flushBuffer();
+        }
+    }
+
+    private static class JsonpHttpServletRequestWrapper extends HttpServletRequestWrapper {
+
+        private static class ByteArrayServletInputStream extends ServletInputStream {
+
```
laserval: I don't like the nested internal classes. I'd prefer having all the internal classes here in a separate package.

laserval:
In the future, when updating to upstream, use `git pull --rebase`. This avoids the merge commit, and also helps you see how your changes relate to the changes on upstream. It also makes more sense for pull requests to be only additional commits to merge into master, without the back-and-forth merge commits.

admin-service/src/main/java/com/findwise/hydra/admin/rest/JsonpCallbackFilter.java
Nycander: new package with all of the internal classes as package-private, and this class as public

jensbengtsson:
To be able to support posting in libraries cross domain CORS would need to be implemented right? So a GUI that uses JSONP would only be able to retrieve data, not add?

ebbesson:
Yes, that's correct. Not every browser supports CORS though e.g. IE<10 . 
CORS can be supported in a similar fashion using e.g http://software.dzhuvinov.com/cors-filter.html
Since CORS-artifact is distributed though maven central it should be fine to add it just like a dependency and configure the web-xml

laserval:
Merged this manually, closing. Nice stuff!

Issue #263, opened 2013-10-25
AbstractProcessStage should have a test for when processing can't be cancelled


laserval:
Test for:
A stage process() method never returns and a `processingTimeout` is set. The processing can't be cancelled, so the stage should destroy its JVM.

There is a TODO in `AbstractProcessStageTest` that this is not tested.

There is an interface `StageKiller` that can be used to inject a non-JVM-killing mock to use for this test.

laserval:
Closed by #286

PR #264, opened 2013-10-25
AbstractProcessStage timeouts and cleaner shutdown

### LINKED ISSUE #247: It is not possible to safely shutdown Hydra

laserval:
This is a continuation of @ssimon's work in #250, which fixes #247, #248 and #249.

Processing stages now create a separate processing thread, which can have an optional timeout set. If the processing takes longer than the timeout, the processing thread will be cancelled. If the thread can't cancel, the stage will destroy the JVM to prevent memory leaks and avoid corrupt state. Core should handle this as any other stage crash and restart the stage.

The motivation for this is that there is no way of knowing if a thread will exit if it can't be cancelled, which means it will take up memory and possibly continue doing work. To protect the rest of the stage JVM, the only way to stop the thread is to destroy the whole JVM.

The tests for `AbstractProcessStage` are now improved and no longer rely on timing to pass.

I couldn't come up with a good test for one case, so I created #263 for that.

### DIFF
--- a/api/src/main/java/com/findwise/hydra/stage/AbstractProcessStage.java
+++ b/api/src/main/java/com/findwise/hydra/stage/AbstractProcessStage.java
@@ -2,18 +2,25 @@ package com.findwise.hydra.stage;
 
 import java.io.IOException;
 import java.util.Map;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import org.apache.http.ParseException;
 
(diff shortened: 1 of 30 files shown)


remen:
Tests are failing in Travis. Any idea why? Don't want to look to carefully on this until it works. Also, the number of commits here is a bit daunting to code review, especially since a few of them are rewrites of previous ones. Would it make sense to rebase/squash?

laserval:
Looking into the test failures.

Good idea about squashing, I'll see what can be done.

laserval:
So, apparently the tests failed....randomly? I'll trigger a few more builds and see if there's any pattern to this.

laserval:
I've now rebased, squashed and reordered some of the commits to make a cleaner history. I replaced the ProcessThread commit with my Future-implementation, but preserved most of the other changes in separate commits. Hopefully it's easier to follow the changes now.

Issue #268, opened 2013-10-28
Several instances of a stage can simultaneously fetch the same document when using MemoryCache


laserval:
When several instances of the same stage are querying a Hydra Core with the MemoryCache enabled, it may happen that a document gets sent to more than one instance.

This is due to MemoryCache not synchronising getting documents from the cache and marking them as fetched.

Symptoms of this happening are  "ERROR internal - Node gave an unexpected respones: 404 not found" errors when documents are saved.

_Note:_ this is not the same issue as #226. That issue is an effect of the Hydra architecture, while this is a concurrency bug.

PR #270, opened 2013-10-30
Fixed concurency issue #268 in MemoryCache

staunsbjerg:
- Added synchronization on critical sections
- Added test case for getAndTag in concurrent environment

### DIFF
--- a/database/src/main/java/com/findwise/hydra/MemoryCache.java
+++ b/database/src/main/java/com/findwise/hydra/MemoryCache.java
@@ -11,12 +11,16 @@ import java.util.concurrent.ConcurrentHashMap;
 
 public class MemoryCache<T extends DatabaseType> implements Cache<T> {
 
-	private ConcurrentHashMap<DocumentID<T>, DatabaseDocument<T>> map;
-	private ConcurrentHashMap<DocumentID<T>, Long> lastTouched;
+	private Map<DocumentID<T>, DatabaseDocument<T>> map;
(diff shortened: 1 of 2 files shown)


remen:
This is one of those cases where the code fix is obviously correct (or at least more correct than what was there before), but where it is very difficult to see whether the test code is correct. My question is, is it worth it? The test is 
- complex,
- tightly coupled to the specific implementation, and
- intrusive (setUnitTestMode).

If you feel strongly that you want to keep it; I'll merge this in. But I vote for removal of the test.

staunsbjerg:
@remen I don't have any strong opinions on keeping the test case or not. I do however feel that intrusive tests with setUnitTestMode is often the best way to go about injecting mocks - else you can end up in constructor overloading-hell.I.e: the code becomes transparent and testable.

I don't think we can make an async test case simplified. So it is up to you guys

remen:
I agree in that I don't think the async test case can be simplified either; and since I don't think it adds enough value in comparison to its complexity, I'll remove it. Thanks for feedback and contribution!

remen:
Manually merged in commit 4dcc1aefa76a39a0e4b985496988caec17daf651

remen:
By the way, I've been down this road before: https://github.com/remen/fork_break, but I've since reached the conclusion that it's not worth the effort. For Java, this seems to be a good lib: http://www.cs.umd.edu/projects/PL/multithreadedtc/overview.html

PR #273, opened 2013-11-19
Adds HttpFetcher and AbstractHttpFetchingProcessStage

Nycander:
Backporting code that @laserval built in a custom stageproject.

I imagine that you would need http fetching capabilities in many stage implementations and I think it should be a part of the Hydra artifacts.

Not sure which though, I chose api since it is the closest match.


Nycander:
Anything I can do to speed this up?

remen:
Hey @Nycander,

Sorry for the delay, and thanks for the contribution! Could you have a chat with @laserval in the office? I think that is a more efficient way than doing it over github.

My remarks:
- Move it to somewhere in stages/processing. Either merge it with stages/processing/basic or stages/processing/web. Ask @laserval :)
- I would prefer if you rebase/squash all of the commits. Fixups of previous commits only makes my job as a reviewer harder.
- There are some missing newlines at the end of some source files. Please add them :P

remen:
Although now I realize that you also want for people to be able to _use_ your libraries as parts of their own stages. That's a tougher nut. I felt the same with regards to e.g. Tika, where I wanted to _use_ Tika in my own stages (actually, I tend to use a single big stage) without using the actual Tika stage.

Dunno. I think talking to @laserval about the overall philosophy is necessary.

laserval:
As you mentioned, I think merging with stages (basic or web) will introduce some odd dependencies - we really shouldn't encourage depending on stage libraries as that will mess up the classpath. I'm not sure where it should be, though - the Abstract stage makes sense to keep in API along with the others (e.g. `AbstractMappingProcessStage`).

Perhaps we need a base library in `stages/processing` that does not contain `@Stage`s?

remen:
A new toplevel module called `utils`

`com.findwise.hydra.utils.HttpFetcher`, or
`com.findwise.hydra.stages.utils.HttpFetcher`

Either works for me.

Nycander:
Wouldn't a toplevel module called "utils" risk getting a really huge list of dependencies?
(Imagine adding utilities for sql, amqp, tika and a bunch of others)

remen:
Fair enough, although my gut feeling is never to feel sorry for computers, only programmers. Code size has never been that big of an issue for me, but I've heard some horror stories from the VPN-trenches :D

`stages/utils/http`
`stages/utils/tika`
`stages/utils/amqp`

ok?

Nycander:
Yeah, I dont really feel sorry for the computer in this case.

I'm afraid of the huge mountain of transitive dependencies, which will most likely collide with each other and create a huge headache for any library user :)

remen:
Good point

remen:
Are we in agreement? Move the reusable stuff to `stages/utils/http`? @laserval?

laserval:
That sounds good to me.

remen:
Any progress?

Nycander:
Yeah, sorry about that. I was on vacation for a week and is now catching up.
Will try to patch this up today.

Nycander:
There you go :)

PR #275, opened 2013-11-22
Adds utility methods for getting and appending to content fields

### LINKED ISSUE #209: LocalDocument should support appending values to fields
Currently, appending values to a field involves:
- Checking if their is such a field
- Checking the type of the field
- Typecasting the value of the field
- Creating a list if the field is not already a list
- Adding the new value of the field

Supporting an append of some sort would make it easier to write stages.

laserval:
Adds getters for the Hydra serializable types (`long`, `double`, `String`, `List` and `Map`, see `api/readme.md`).
Adds method for appending values to content fields without type-checking, since doing that kind of check reliably is difficult at runtime (there will always be cases where the type is unknown). Additionally, generic types of lists and maps are erased on serialization, so all documents received in a stage will have maps and lists with generic type `Object`.

This resolves #209

There might be more utility methods that could be useful; a quick check through the basic stages show a frequent pattern of `if string, do this and output a string; if list, do something else and output a list`. These methods don't really help with that.

### DIFF
--- a/api/readme.md
+++ b/api/readme.md
@@ -1,0 +1,65 @@
+# Hydra API
+
+For anything shared between stages and other components, the hydra-api package is used as a common base.
+
+
+### RemotePipeline
+
+Interaction with an instance of hydra-core occurs through the RemotePipeline class, and that class only. Some details about data transfer via the RemotePipeline class are worth mentioning, because they affect the way you handle anything that has passed through this class.
+
+When serializing and deserializing a LocalDocument or LocalQuery some limitations/generalizations are imposed by the Json nature of the serialization:
+
+**Arrays, Lists and Sets:** Arrays of any type and Lists and Sets of any generic type become List<Object> (with ArrayList as the implementing class)
+
+**All integer-types:** All Integers, be they of type int, long, byte or short, become long or int depending on size
+
+**All floating-point types:** All floating point types, be they float or double become double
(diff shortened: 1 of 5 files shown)


laserval:
I've changed to throw an `IncorrectFieldTypeException` from methods where that makes sense, and to differentiate from the empty field state.

I also rebased on the merged changes, and split the formatting changes to a separate commit.

laserval:
Updated with changes to the long and double methods, to throw separate exceptions for empty field and wrong type.
Reasoning for throwing exceptions is ease of use - you wouldn't expect a getter for a number to return null.

laserval:
Yes.

I believe in this case you will almost always want to handle the fact that the fields are of the wrong type. I doubt that it is reasonable for a stage to crash on an uncaught exception due to a field being the wrong type, so it makes sense to encourage developers to handle the cases where the field is not what they expected.

As for the `FieldIsEmptyException`, it makes it obvious that the method behaves differently from the other getters.

remen:
> I doubt that it is reasonable for a stage to crash on an uncaught exception due to a field being the wrong type

Agreed. But I would have implemented that in some other fashion. Possibly by inheriting from ProcessException, or by something which is caught by the main loop and turned into a ProcessException.

But this is a matter of taste only. Merging.

Issue #240, opened 2013-08-22
Documents should have getter for attached files


laserval:
Currently, getting hold of attachments requires:
1. Getting hold of the RemotePipeline from the current stage
2. Calling `getFileNames` to find out what attachments, if any, the document has
3. Calling `getFile` for reach file you want
4. Getting the inputstream from the returned DocumentFile
5. Streaming the file to wherever it is needed

laserval:
There is a problem here, in that documents don't contain the attachment. To fetch an attachment, a request must be made to Core to get the content.

A utility method could be added to `RemotePipeline` that simplifies this to a single step `getFiles` that returns all attachments as `DocumentFile`s.

To get an actual attachment-getter for LocalDocument, something more needs to change.

PR #276, opened 2013-11-25
Adds method for retrieving all document attachments, changes Date serialization

laserval:
_NOTE_: This PR changes the serialization format of `Date` objects to include milliseconds. The old format is added as a legacy format that `SerializationUtils` can deserialize, but not serialize to. This affects all configuration (stage, stagegroup) as well as document serialization.
Previously, sub-second components were lost in serialization, which made tests difficult to write and required weird workarounds in test code.

Legacy format: `2013-01-13T14:02:56+0100`
New format: `2013-01-13T14:02:56.123+0100`

**Adds `getFiles` to `RemotePipeline`, for fetching all attachments to a document**. This addresses #240, by making it easier to get the attachments.

I also introduced the use of WireMock for testing `RemotePipeline`, something that should be expanded to all methods in the class.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -27,6 +27,11 @@
 			<scope>compile</scope>
 		</dependency>
 		<dependency>
+			<groupId>ch.qos.logback</groupId>
+			<artifactId>logback-classic</artifactId>
+			<version>1.0.10</version>
+		</dependency>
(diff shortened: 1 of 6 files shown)


remen:
Looks good. Merging.

PR #277, opened 2013-11-25
Adds more generic exception to stages' init() method

### LINKED ISSUE #189: AbstractStage.init() should throw more generic exception.
It is possible that one might want to refuse a stage to start other than missing required arguments, such as validating that an integrating system is unavailable or that authentication is invalid.

laserval:
Fixes #189

`init()` now declares a second, more generic exception (`InitFailedException`) for when initialization of a stage fails due to factors other than missing or incorrect arguments (such as connection failure to external service).

Changes output stages to use the new exception.


Nycander:
Nice!

remen:
Ouch. The travis build failed on the new tests dealing with processing timeouts. That is _not_ good. Created #279.

remen:
Tested and merged manually. Tests are failing randomly, but not due to these commits.

PR #282, opened 2013-11-27
Continue instead of return in RegexStage

andreassalomonsson:
The regex stage will now continue with the rest of the configurations, instead of returning, when a document is missing the input field for one of the configurations.


Nycander:
Looks good :+1:

PR #283, opened 2013-11-29
Adds check for empty configuration to Admin Service stage config endpoint

### LINKED ISSUE #245: Posting a stage configuration with an attachment (e.g. library file) removes properties
If a stage configuration is POSTed to `libraries/{library}/stages/{stage}` with a properties map as body and an attached file (e.g. library jar), the properties of the stage will be set  to `{ "" : null }`.

This can happen when e.g. using a REST helper tool to create requests, and forgetting to remove an attached library file when setting up stages for that library.

The Admin service should either reject the request or ignore the attached file.

laserval:
The Admin Service will now throw an error if the supplied configuration is empty or missing the required `stageClass` parameter. This mostly fixes #245 as the config won't be written to the database, although you will get a syntax error instead - not entirely sure why the attached file breaks the request in that manner.

The Admin Service will now also respond with an error message body of its own when encountering problems, instead of falling back to the container's error page.

For example, a syntax error in the supplied JSON is now a client error, not a server error.

### DIFF
--- a/admin-service/src/main/java/com/findwise/hydra/admin/rest/ConfigurationController.java
+++ b/admin-service/src/main/java/com/findwise/hydra/admin/rest/ConfigurationController.java
@@ -1,13 +1,17 @@
 package com.findwise.hydra.admin.rest;
 
 import java.io.IOException;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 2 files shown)


remen:
Fix the DRY violation and have a look at the double exception thing. Update request when fixed and I'll merge.

PR #284, opened 2013-11-29
Fixup in AbstractStage.getInstances

remen:


remen:
I'm gonna submit a bigger patch with refactoring of command line argument handling and stage startup.

PR #285, opened 2013-11-30
Fixups of some DRY violations when parsing command line arguments

remen:
Hi,

This is something I've wanted to do for a while, but finally got around to while trying to understand how StageGroups work. I noticed that the logic for parsing command line arguments was duplicated and not clearly defined. This is an attempt to fix that. I also took the opportunity to extract some stuff from AbstractStage (which is becoming a bit big imho).
- Created StageCommandLineArguments class for parsing String[] args
- Moved logic for stage creation from AbstractStage into a StageFactory
- Other misc. refactoring + additional comments

### DIFF
--- a/api/src/main/java/com/findwise/hydra/local/RemotePipeline.java
+++ b/api/src/main/java/com/findwise/hydra/local/RemotePipeline.java
@@ -45,7 +45,8 @@ public class RemotePipeline {
 	public static final String FILENAME_PARAM = "filename";
 
 	public static final int DEFAULT_PORT = 12001;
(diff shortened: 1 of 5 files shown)


remen:
We can wait with merging this pull request to after the 0.4.0 release. I want to do more here, but don't want to introduce the risk at this stage.

laserval:
Do you still want this merged before the API refactoring?

remen:
Nah, under the assumption that we _will_ be merging the API refactoring, I don't think we need this pull request separately. Closing.

PR #288, opened 2013-12-10
Rewrite of test-suite

remen:
There was a submodule called "test-suite" containing a test of a pipeline. Great! However:
- It wasn't part of the parent pom build
- It downloaded and ran its own version of mongodb, which doesn't work on travis
- It didn't upload jar-with-dependencies to the core. Instead it ran stages "inline" (afaict)

I needed a full scale functional test suite while doing the refactoring, so I decided to rewrite it.


remen:
Sorry about the `integration_tests` commit. Should have rebased before creating the pull request. Please let me know if you want me to redo it or if it's ok.

laserval:
Good stuff, although I believe it does something different than the old `test-suite`. That was built to test pipelines (you modify it with your pipeline). I haven't used it, but maybe @ssimon has. I think this is more useful to us though.

test-suite/functional-tests/src/test/java/com/findwise/hydra/FullScaleIT.java
```diff
@@ -0,0 +1,197 @@
...
+public class FullScaleIT {
+	Logger logger = LoggerFactory.getLogger(FullScaleIT.class);
+	MongoConfiguration mongoConfiguration;
+	MongoConnector mongoConnector;
+	private Main core;
+
+	@Before
+	public void setUp() throws Exception {
+		mongoConfiguration = new MongoConfiguration();
+		mongoConnector = new MongoConnector(mongoConfiguration);
```
laserval, line 48: This test should use a non-default pipeline, like the integration tests in Core and the Mongo implementation.

laserval:
I've worked a bit with this now, and it's an excellent tool! We should add functional tests for some different scenarios to ensure we catch runtime-functionality. I suppose there are ways to improve it, but I think this is a great start.

I've created a pull request to your branch with an additional test (to catch that init() is run), as well as a bugfix and a change to use a non-default pipeline/namespace for the test.

remen:
I'm glad you like it. I think these types of tests are the ones that add the most value (of course you should _also_ write unit tests) and are most resilient to refactoring. I'll have a look at your changes, and might add some of my own  since I think I'm using stage groups wrong.

remen:
Currently (because I wasn't thinking about it) there is one stage-group per stage. I don't really know what the "canonical" way of using Hydra is. What does the command line inserter do per default? What should we be testing?

laserval:
> What does the command line inserter do per default? What should we be testing?

The default is one group per stage. The only current benefit to stage groups is that you cut down on JVM overhead.
I think we should at least have both cases - i.e. we should have both a stage group with several stages and one with a single stage.

remen:
Good idea, I'll get right on it.

remen:
Done. The change was quite invasive, but I think it makes it easier to understand the overall logic anyway.

PR #296, opened 2014-01-02
Add a basic stage for converting dates to iso8601Z date (zulu date)

leonardsaers:
Solr consumes dates strictly formatted as iso8601 with time zone +0. The format is sometimes called zulu time. This stage formats dates into ISO8601 zulu format. So far it handles epoch time and ISO8601 with different time zones.

Is this stage worth adding to Hydras basic stages?

### DIFF
--- a/stages/processing/basic/src/main/java/com/findwise/hydra/stage/ZuluDateFormatter.java
+++ b/stages/processing/basic/src/main/java/com/findwise/hydra/stage/ZuluDateFormatter.java
@@ -1,0 +1,118 @@
+package com.findwise.hydra.stage;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
+
+import com.findwise.hydra.local.LocalDocument;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.regex.Pattern;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+import org.joda.time.format.DateTimeFormat;
+import org.joda.time.format.DateTimeFormatter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Converts epoch time and ISO 8601 with different time zones to Zulu time. 
+ * Zulu time is ISO 8601 where the time zone is +0. The format has the following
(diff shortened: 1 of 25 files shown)


stages/processing/basic/src/main/java/com/findwise/hydra/stage/ZuluDateFormatter.java
```diff
@@ -0,0 +1,118 @@
...
+                
+                if ((contentField.getValue() instanceof String)) {
+                                
+                    parseDate(contentField, doc);
+                }
+            }
+        }
+    }
+
+    private void parseDate
```
laserval, line 86: Weird formatting

laserval:
I think a more generic implementation makes more sense to add. This is a good start for that, but it has too specific a name and purpose - why not simply `DateFormatterStage`? Then Zulu date can be a default, to start with.

leonardsaers:
I agree that such a stage would be better. But it would add more complexity to the stage, converting from any format to any format.

To start with, should we just remain ZuluDateFormatter to DateFormatterStage?

stages/processing/basic/src/main/java/com/findwise/hydra/stage/ZuluDateFormatter.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    private String fromISO8601ToZuluDate(String date) {
+        DateTimeFormatter formatter = 
+                DateTimeFormat.forPattern("yyyy-MM-dd'T'HH:mm:ssZ");
+        DateTime d = formatter.parseDateTime(date).withZone(DateTimeZone.UTC);
+        return d.toString(DateTimeFormat.forPattern("yyyy-MM-dd'T'HH:mm:ss'Z"));
+    }
+
+    private String fromEpocToZuluDate(String epoc) {
```
laserval, line 109: Should be _epoch_

laserval:
> To start with, should we just remain ZuluDateFormatter to DateFormatterStage?

Yes, that's a good start. The configuration doesn't look Zulu-specific, so it shouldn't be a problem to make the stage more generic in the future.

leonardsaers:
Class name is now changed to DateFormatterStage

laserval:
Closing in favor of merge of PR #308 , which is just a rebase of this to solve some merging issues.

PR #300, opened 2014-01-16
HttpFetcher shouldn't consume the HttpEntity

andreassalomonsson:
HttpFetcher no longer consumes the HttpEntity making it possible to actually read the content stream.


stages/utils/http/src/main/java/com/findwise/utils/http/HttpFetcher.java
```diff
@@ -122,12 +121,7 @@ public HttpEntity fetch(String identifier,
                 attempts++;
                 StatusLine status = response.getStatusLine();
                 if (HttpStatus.SC_OK == status.getStatusCode()) {
-                    HttpEntity entity = response.getEntity();
-                    try {
-                        return entity;
-                    } finally {
-                        EntityUtils.consume(entity);
-                    }
+                    return response.getEntity();
```
laserval, line 127: This means the caller must consume the entity, should that be added to the method or class javadoc?

laserval:
This fails to build. Could be duplicate classes due to dependencies, please check the build logs.

PR #304, opened 2014-01-24
Fixes "socket closed" bug and failing HTTPutils test

### LINKED ISSUE #302: HttpFetcherTest fails randomly
The `HttpFetcher` tests fail randomly. It seems like the connection to WireMock closes suddenly.

laserval:
A request entity was returned by `httputils.fetch()`, then a finally block reset the request. This caused unpredictable behaviour with "Socket Closed" thrown by anything that read the entity body.

This also fixes the test failing on some platforms and machines (such as mine, and the Findwise Jenkins build server).

Fixes #302

### DIFF
--- a/stages/processing/basic/pom.xml
+++ b/stages/processing/basic/pom.xml
@@ -29,6 +29,11 @@
 			<version>${project.version}</version>
 		</dependency>
 		<dependency>
+			<groupId>commons-io</groupId>
+			<artifactId>commons-io</artifactId>
+			<version>2.4</version>
+		</dependency>
+		<dependency>
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
 			<version>4.11</version>
--- a/stages/processing/basic/src/main/java/com/findwise/hydra/stage/AbstractHttpFetchingProcessStage.java
+++ b/stages/processing/basic/src/main/java/com/findwise/hydra/stage/AbstractHttpFetchingProcessStage.java
@@ -1,5 +1,6 @@
 package com.findwise.hydra.stage;
 
+import com.findwise.hydra.local.IncorrectFieldTypeException;
 import com.findwise.hydra.local.LocalDocument;
 import com.findwise.utils.http.HttpFetchConfiguration;
 import com.findwise.utils.http.HttpFetchConfigurationBuilder;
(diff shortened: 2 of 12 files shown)


laserval:
I'm not sure how to test this, as the problem was that there was a return statement in a `try { ... } finally` block. The tests that failed would do so randomly. The only way I can think of to produce a failing test is to call the method repeatedly.

I see what you mean with the `BAD_REQUEST`, and it's a stage-specific feature that's sneaked in. This code comes from a project where we had problems with the request object getting corrupted when reused.
Perhaps just creating a new request object for each call is better.

remen:
> I'm not sure how to test this, as the problem was that there was a return statement in a `try { ... } finally` block. The tests that failed would do so randomly. The only way I can think of to produce a failing test is to call the method repeatedly.

Ok, no problem. Just wanted to see what the error condition was that would trigger it since I cannot immediately understand the logic. NB: if testing is hard, it might be a code smell.

> I see what you mean with the `BAD_REQUEST`, and it's a stage-specific feature that's sneaked in. This code comes from a project where we had problems with the request object getting corrupted when reused.
> Perhaps just creating a new request object for each call is better.

Well, at least increase the number of attempts so that it doesn't end up in an infinite loop! That's what's scaring me at the moment.

As a side note, the fact that stage-specific features sneak in could be a symptom of us adding stages to the hydra distribution too quickly. Could be that we should be more conservative, since adding stuff comes at such a high cost.

laserval:
I've reworked this a bit, since I realised the tests for `AbstractFetchingProcessStage` were checking `HttpFetcher` functionality, and there was some very odd mocking going on. Additionally, there was no implementation using the abstract stage.

I've also removed the special case for `BAD_REQUEST`, and added a `RequestProvider` interface that should allow stages to also run `POST` or whatever other HTTP request they'd want.

remen:
I have some proposed changes incoming in a pull request. Other than that, I would have considered using some sort of callback mechanism to deal with retries, error handling and closing of resources instead of abstract classes.

I'm a bit uncertain about the use case for letting the stage provide a _different_ URIs between attempts, but I guess you have a specific case in mind.

laserval:
I've merged your changes.

The idea with different URIs is as a kind of fallback system. I suppose that could be better done by different implementations of the URI provider. There is too much functionality in the abstract class, I agree.

But perhaps this is enough for the issue that was the target of this PR. I'd be willing to go over this before it makes it into a release.

remen:
Excellent. Merging.

remen:
(damn it, this means I probably have to rebase the API stuff yet again :smile:)

PR #305, opened 2014-01-28
API rewrite

remen:
Big rewrite of the API as announced on the mailing list.
- Removed all input stages
- Stages no longer have access to `RemotePipeline`.
  - To discard a document, call `LocalDocument#discard`
  - To get document attachments, call `LocalDocument#getFile` et.al.
- Now uses Guava Service instead of handling the lifecycle of long living threads ourselves.
- ~~Uses Jackson for json deserialization. This removes the need for our own deserialization library.~~
- ~~Uses hibernate-validation for validation of Stage beans.~~
- All exceptions thrown in stage implementations are now caught and fails the document.


laserval:
Awesome! Nice with the squashed commits, will look through this.

api/src/main/java/com/findwise/hydra/stage/ProcessStageRunner.java
```diff
@@ -0,0 +1,128 @@
...
+		} else {
+			future.get();
+		}
+	}
+
+	private void persist(LocalDocument doc) throws IOException, JsonException {
+		logger.debug("Saving document to RemotePipeline..");
+		if(!remotePipeline.save(doc)) {
+			// TODO: Why are we cloning here?
+			LocalDocument ld = new LocalDocument(doc);
```
laserval: Wondering the same thing about this line. It was introduced in 5dc4438fbdcc7ad6015a9d7574e00072b0534794. I can only think of some kind of concurrency issue. With `RemotePipeline` now being immutable, I can't see what cloning would achieve.

remen:
Added some fixes to your comments. Once you've reviewed, I'll rebase and squash them into the big commit.

api/src/main/java/com/findwise/hydra/local/LocalDocument.java
```diff
@@ -10,15 +10,14 @@
 import java.util.Set;
 import java.util.HashMap;
 
-import com.findwise.hydra.Document;
-import com.findwise.hydra.DocumentID;
-import com.findwise.hydra.JsonException;
-import com.findwise.hydra.SerializationUtils;
+import com.findwise.hydra.*;
```
Nycander: You should never use \* for imports, please fix the IDE :)

api/src/main/java/com/findwise/hydra/local/RemotePipeline.java
```diff
@@ -10,21 +10,21 @@
 import java.util.List;
 import java.util.Map;
 
+import com.findwise.hydra.*;
```
Nycander: Same as above, fix the IDE and remove all \* imports

api/src/main/java/com/findwise/hydra/local/RemotePipeline.java
```diff
@@ -45,24 +45,23 @@
 	public static final String FILENAME_PARAM = "filename";
 
 	public static final int DEFAULT_PORT = 12001;
-	public static final String DEFAULT_HOST = "127.0.0.1";
+	public static final String DEFAULT_HOST = "localhost";
+	public static final int DEFAULT_LOG_PORT = 12002;
```
Nycander, line 53: I have no opinion, just mainly curious, why localhost instead of 127.0.0.1?

api/src/main/java/com/findwise/hydra/stage/AbstractProcessStage.java
```diff
@@ -1,217 +1,44 @@
...
+import java.util.concurrent.TimeUnit;
 
-/**
- * Base class for all stages that do processing of documents in Hydra.
- * 
- * @author anton.hagerstrand
- * @author simon.stenstrom
- * @author anders.rask
- * @author joel.westberg
- *
```
Nycander, line 29: I like the fact that there is less code in these abstract classes! :+1:

api/src/main/java/com/findwise/hydra/stage/AbstractStage.java
```diff
@@ -1,362 +0,0 @@
-package com.findwise.hydra.stage;
```
Nycander, line 1: Wooo!

api/src/main/java/com/findwise/hydra/stage/ProcessStageRunner.java
```diff
@@ -0,0 +1,126 @@
+package com.findwise.hydra.stage;
+
+import com.findwise.hydra.JsonException;
+import com.findwise.hydra.local.LocalDocument;
+import com.findwise.hydra.local.RemotePipeline;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.util.concurrent.*;
```
Nycander: :(

api/src/main/java/com/findwise/hydra/local/LocalDocument.java
remen: Fair enough. Is it ok for static imports, e.g.

``` java
import static org.junit.Matchers.*;
```

api/src/main/java/com/findwise/hydra/local/RemotePipeline.java
remen, line 53: Very good question. Don't remember now. I'll revert it; the fewer changes the better.

stages/out/elasticsearch-out/src/main/java/com/findwise/hydra/output/elasticsearch/ElasticsearchOutputStage.java
```diff
@@ -1,10 +1,9 @@
 package com.findwise.hydra.output.elasticsearch;
 
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
 
-import com.findwise.hydra.stage.InitFailedException;
+import com.findwise.hydra.stage.*;
```
Nycander: :(

api/src/main/java/com/findwise/hydra/stage/ProcessStageRunner.java
remen: I'll do a full search replace for all occurrences.

Nycander:
More removed code than added, achievement unlocked, grats :D

remen:
Let's see if it stays that way after expanding all the wildcard imports :smile:

api/src/main/java/com/findwise/hydra/local/LocalDocument.java
Nycander: It's not as harmful as with normal imports. But for consistency I usually
stick to explicit imports for everything...

remen: FYI: Clean code disagrees: http://stackoverflow.com/a/1992695

Personally, I don't have a preference, except I want to have a consistent style so we don't have to have these conversations all the time :smile:

ssimon: Two points, of which none apply when you're using a decent IDE though..

Nycander: Hard dependencies are good. Long import lists are hidden by IDE's by default and is very useful when reading other peoples code. Also less likely to break when people add classes to packages you wildcard import.

In the companies I've worked at have always forbidden wildcard imports. Google also has it as a rule in their style guide for java:
http://google-styleguide.googlecode.com/svn/trunk/javaguide.html#s3.3.1-wildcard-imports

remen: Very good points. I don't care about imports since I don't even see them (and let the IDE import them automatically). Therefore, the style with the least amount of potential problems is the best; which is hard deps.

I'm convinced. Thanks!

remen: Any opinions on import order/grouping? Can/should I use the google style? It would help me if I could just run "optimize imports" in IDEA and know that it will be up to snuff.

[UPDATE] My personal preference is actually the reverse of the google style. I.e. I prefer:

``` java
import java.util.Map;
import java.util.List;

import org.slf4j.LoggerFactory;
import org.slf4j.Logger;

import com.findwise.hydra.api.LocalDocument;
import com.findwise.hydra.api.RemotePipeLine;

import static com.google.guava.Preconditions.checkNotNull;
```

The reason being that I tend to read import statements bottom-up, and would like to get most relevant info first.

Nycander: I don't really have an opinion on the order.

Standards are good, so as long as there is some kind of decision and uniformity - I'm happy :)

remen: Ok. I'm deciding then. We'll use the google style for import ordering. That way, we can simply refer to an online document and also reuse (big parts of) their prewritten style configurations for the IDEs.

Nycander: :+1:

ssimon: Should we also change to space instead of tabs as suggested? We've already
changed the prefered style for other components at FW, so it would be nice
not to switch (even though I prefer tabs)...

remen: Has SE/SDP created a style guide?

laserval: Why not adopt the Google style document entirely for Hydra? There are formatting files ready for it, right? That way we have a clear style to point to. We don't have to do any gigantic conversion right away, but we could use it to get a coherent style to all files we modify.

ssimon: Kind of.. :)

"Code style guidelines
- Indentation of code is four spaces, no tabs
- If and while statements should contain surrounding curly braces

"
from the jellyfish project on bitbucket..

ssimon: [?]

remen: Ok, well, SE/SDP will get together and develop a style guide together with settings for all major IDEs. It is my guess that we will be using the google style, but with four spaces instead of two (however, we might want to use two just to conform as much as possible to google).

remen: @laserval I propose we delay that decision for Hydra until SE/SDP has decided for findwise overall.

However, I will be optimizing my imports according to the google style and remove all wildcard imports.

laserval: Sounds good. I've started the discussion internally.

laserval:
This looks good! Rebasing the changes to `getParameters()` so that it's clear that the logic wasn't changed would be best.

api/src/main/java/com/findwise/hydra/local/RemotePipeline.java
remen, line 53: Now I know why, it's because the field was being ignored by GroupStarter where it was instead  defaulting to using localhost. I made sure that the cmdline args parser respects the default, but wanted to make sure that the change preserved the _current_ behavior!

[UPDATE]: So I _won't_ be reverting it.

remen:
Ok, it's been rebased against current master, and all wildcard imports have been removed.

laserval:
I'm thinking we could soon go ahead and merge this (I suppose it needs another rebase, unfortunately, since the recent PR merges).
Should we do a 0.5.0 release after that? I'm not sure if we have any bug fixes to 0.4.0 currently.

laserval:
I've tried a rebase, and the conflicts are all related to the newly added stages `DateFormatterStage` and `SimpleHttpFetchingStage`, as well as the changes in `AbstractHttpFetchingStage`. All `ProcessException` and `setParameters()` references need to be removed and some extra setters added.

remen:
Release early, release often. I say do it.

Do you want me to do the rebase, or code review yours?

laserval:
I can put up my rebased branch.

laserval:
I've created PR #313 which contains the rebase.

laserval:
I've now merged #313 which contained this rebase and handled the changes from the PRs that were merged while reviewing. Great job, @remen! I plan to do a release of 0.4.1 (GUI+stages on 0.4.x API) and 0.5.0 (GUI+stages+new API) this week. Hopefully I'll have time for both :)

PR #310, opened 2014-02-13
Admin service gui

ebbesson:
This pull request contains a simple UI for adding and configuring stages.

Currently it can only add and update one stage per stage-group.

Changes to the Admin Service REST API:
- `stagegroups` now does library files inspection, meaning it will take the same amount of time (or more) than the `libraries` endpoint to return. On the other hand, you get the description, required-flag and current value for all properties.

0-admingui-stagegroups
1-admingui-stagegroups-modify-stage
2-admingui-libraries
3-admingui-libraries-add-stage


laserval:
Should we replace the JS libraries with minified versions? Confused me a bit when the size of the PR is 13000 lines.

remen:
Is it possible to add as dependencies in some eay?

admin-service/pom.xml
```diff
@@ -13,19 +13,22 @@
...
-            </plugin>
-        </plugins>
-    </build>
+	<build>
+		<finalName>hydra</finalName>
+		<plugins>
+			<plugin>
+				<groupId>org.eclipse.jetty</groupId>
+				<artifactId>jetty-maven-plugin</artifactId>
+				<version>9.0.5.v20130815</version>
```
laserval: This won't run on Java 6. Downgrade to previous version, such as `8.1.14.v20131031` (need to change group then, to `org.mortbay.jetty`).

laserval:
I'm not sure. Perhaps using this: http://www.webjars.org/

> Is it possible to add as dependencies in some eay?
>

admin-service/src/main/java/com/findwise/hydra/admin/rest/DatabaseConfig.java
```diff
@@ -7,7 +7,7 @@
 public class DatabaseConfig implements DatabaseConfiguration {
```
laserval, line 8: Hm, this should use the defaults from the `database` package. I wonder if the Spring stuff can still be used in a nice manner then.

admin-service/src/main/java/com/findwise/hydra/admin/rest/WebConfig.java
```diff
@@ -2,9 +2,16 @@
...
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurerAdapter;
 
 @Configuration
 @EnableWebMvc
-public class WebConfig {
+public class WebConfig extends WebMvcConfigurerAdapter {
 
+	@Override
+	public void addResourceHandlers(ResourceHandlerRegistry registry) {
+		registry.addResourceHandler("/gui/**").addResourceLocations("/WEB-INF/static/");
```
laserval, line 14: When accessing `/gui`, none of the resources can be reached.
The log hints at why:

```
Feb 14, 2014 11:27:38 AM org.springframework.web.servlet.handler.AbstractUrlHandlerMapping registerHandler 
INFO: Mapped URL path [/gui] onto handler '/gui'
Feb 14, 2014 11:27:38 AM org.springframework.web.servlet.handler.AbstractUrlHandlerMapping registerHandler 
INFO: Mapped URL path [/gui/**] onto handler of type [class org.springframework.web.servlet.resource.ResourceHttpRequestHandler]
```

Could it be this resourcehandler mapping?

ebbesson:
The webjars looks like a nice way of handling web-dependencies
I'll see what I can do with them

admin-service/pom.xml
ebbesson: Done

admin-service/src/main/java/com/findwise/hydra/admin/rest/WebConfig.java
ebbesson, line 14: Nice catch, I've changed from relative paths to absolute paths in index.html.

admin-service/src/main/java/com/findwise/hydra/admin/rest/DatabaseConfig.java
laserval, line 8: I "fixed" this in a pull request to your branch, @ebbesson .
We really need to do something about the configuration classes...

laserval:
I think what I'll be doing now is to merge this, then rebase #313 again and merge that. That way we can do a release of Hydra 0.4.1 or so with the GUI and new stages, then release 0.5.0 with the API changes.

PR #312, opened 2014-02-17
Ensures functional tests clean up their database use

laserval:
I noticed the functional tests were leaving behind a database in MongoDB. This should make sure they don't.


remen:
Good call.

PR #322, opened 2014-03-10
Fixes deprecation, Maven, javadoc and unchecked warnings

laserval:
There were a number of compile warnings, especially references to the deprecated `junit.framework.Assert`. I've replaced those with the correct one (`org.junit.Assert`) and fixed a bunch of other warnings.


remen:
Otherwise looks good to go. Answer question above and merge at will.

PR #323, opened 2014-03-13
ADD to stages tika: class documentation for stage SimpleFetchingTikaStag...

leonardsaers:
It took some time to understand how metadata is added to SimpleFetchingTikaStage. So I added some class documentation which should explain how the stage works.


laserval:
Looks good, didn't know about the `{@code ... }` tag, I've always used HTML tags.

PR #326, opened 2014-04-01
Restructure readme and add stage libs to distribution

laserval:
- Readme is better structured with the getting started at the top, instead of compilation instructions.
- Most stage libraries are now found in `distribution/lib` after building the parent pom.

### DIFF
--- a/admin-service/pom.xml
+++ b/admin-service/pom.xml
@@ -16,6 +16,24 @@
 	<build>
 		<finalName>hydra</finalName>
 		<plugins>
+            <plugin>
+                <artifactId>maven-antrun-plugin</artifactId>
+                <version>1.4</version>
+                <executions>
+                    <execution>
+                        <id>copy</id>
+                        <phase>compile</phase>
+                        <configuration>
+                            <tasks>
+                                <copy file="src/main/resources/admin-service.properties" tofile="target/admin-service.properties" />
+                            </tasks>
+                        </configuration>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
 			<plugin>
 				<groupId>org.mortbay.jetty</groupId>
 				<artifactId>jetty-maven-plugin</artifactId>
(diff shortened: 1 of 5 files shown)


distribution/pom.xml
```diff
@@ -41,8 +41,16 @@
 						<configuration>
 							<tasks>
 								<copy file="../core/target/hydra-core-jar-with-dependencies.jar" tofile="bin/hydra-core.jar" />
-								<copy file="../core/target/resource.properties" tofile="bin/resource.properties" />
+								<copy file="../core/target/resource.properties" tofile="bin/hydra-core.properties" />
```
remen, line 92: When adding stuff here, it's a good idea to add the projects as dependencies as well so that the reactor build is guaranteed to be in the correct order.

admin-service/pom.xml
```diff
@@ -13,22 +13,40 @@
...
+            <plugin>
+                <artifactId>maven-antrun-plugin</artifactId>
+                <version>1.4</version>
+                <executions>
+                    <execution>
+                        <id>copy</id>
+                        <phase>compile</phase>
+                        <configuration>
+                            <tasks>
+                                <copy file="src/main/resources/admin-service.properties" tofile="target/admin-service.properties" />
```
remen, line 28: This double copying surprised me, but I'm guessing there's a good reason for it.

remen:
Looking good. Add dependencies to ensure correct reactor ordering and merge.

admin-service/pom.xml
laserval, line 28: I was mostly following the "pattern" from Hydra Core. If you're only building this project, I think it makes sense to add the config file next to the binary.

It is a bit unnecessary, though. Do you think we should remove it from both Core and the admin service?

laserval:
I updated the distribution pom with the dependencies, and added the two other libraries I hadn't added (groovy-runner and debugging). Might as well have all stages in there! Had to add jar-with-dependencies packaging to groovy-runner, which goes to show how much that stage has been used...

Issue #236, opened 2013-08-16
Refactoring of Tika stages


ssimon:
Especially the TikaUtils class is poorly written. The Utils should return the text content (or metadata or whatever) rather than adding it to the document. The stage should be responsible for adding the returned data to the document (or do whatever you may want to do with it!)

PR #327, opened 2014-04-02
Extract Tika util and update to 1.5

laserval:
I've refactored the Tika stages and moved their central parsing util out into a Tika util. I've also split the util into some separate classes and removed all the staticness.

There should now be less duplicated code in the Tika stages.

I've also updated Tika from 1.1 to 1.5

My initial plan was to move the entire dependency on Tika out into the util, including the `tika-parsers` dependency. Unfortunately it proved very difficult, due to the way tika-parsers introduces all these conflicting classes. In the end I let the util have `tika-core` as a dependency, then the stages can provide any necessary parsers.

I'm not sure how we want to use the util projects. I avoided Hydra dependencies now, but in general it could make sense to have them depend on Hydra. That way some of the code in the new classes in tika-stages could be moved to the util.

Related to #236, does some refactoring but we could probably do more. For example, `TextExtractionStage` does mostly the same thing as `SimpleFetchingTikaStage`, but with some extra operations on filenames and formats.


stages/processing/tika/pom.xml
```diff
@@ -1,147 +1,177 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.findwise.hydra</groupId>
+            <artifactId>hydra-tika-utils</artifactId>
+            <version>${project.parent.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.tika</groupId>
+            <artifactId>tika-parsers</artifactId>
+            <version>1.5</version>
```
laserval, line 30: Forgot to show this in a commit message, but I updated Tika from 1.1 to 1.5

PR #346, opened 2014-11-04
0.4.x

ssimon:
Added a split stage, a static action stage and secured the SimpleHttpFetchingStage to handle entities without specified encoding

### DIFF
--- a/stages/processing/basic/src/main/java/com/findwise/hydra/stage/SimpleHttpFetchingStage.java
+++ b/stages/processing/basic/src/main/java/com/findwise/hydra/stage/SimpleHttpFetchingStage.java
@@ -25,6 +25,7 @@ public class SimpleHttpFetchingStage extends AbstractHttpFetchingProcessStage {
 	@Parameter(required = true, description = "Destination field for fetched content")
 	private String outputField;
 
+
 	@Override
 	public URI getUriFromIdentifier(String identifier, int attempts) throws URISyntaxException {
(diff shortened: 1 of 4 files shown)


stages/processing/basic/src/main/java/com/findwise/hydra/stage/SimpleHttpFetchingStage.java
```diff
@@ -34,8 +35,13 @@ public URI getUriFromIdentifier(String identifier, int attempts) throws URISynta
...
 		try {
 			InputStream inputStream = responseEntity.getContent();
-			Charset encoding = Charset.forName(responseEntity.getContentEncoding().getValue());
-			String content = IOUtils.toString(inputStream, encoding);
+			String content;
+			if (responseEntity.getContentEncoding() != null) {
+				Charset encoding = Charset.forName(responseEntity.getContentEncoding().getValue());
+				content = IOUtils.toString(inputStream, encoding);
+			} else {
+				content = IOUtils.toString(inputStream);
```
laserval: Should we be assuming the default encoding here? I'm not sure about that, I would prefer assuming UTF-8.

Nycander: Added a fallback encoding configuration parameter

Nycander:
I think I'm done now.
@laserval @ssimon merge?

PR #352, opened 2014-11-24
Adds Main.stop() and cleans up a few log-strings.

Nycander:


laserval:
Looks good! We could potentially add other service wrapper stop methods or integrations if needed in the future.

PR #362, opened 2015-02-13
Adds configuration option for Core to set default JVM parameters on all stagegroups

laserval:
This makes it easier to configure stagegroups where some stages need e.g. keystore files, or to set a max heapsize for all stages.

Also fixes issues with stages not removing jars when a stage library is renamed (e.g. when versions change).

This also allows the `jvm_parameters` stage parameter to be cross-platform: As long as paths in JVM parameters are quoted properly no special handling for different platforms should be needed. Before this, the JVM parameters had to be quoted inside the string on Windows.


laserval:
Wrong version.

